Issue #3, opened 2017-05-12
fix compilation when including bse header for out-of-tree code


swesterfeld:
SpectMorph 0.3.2 doesn't compile properly agains BEAST master due to include issues

See https://mail.gnome.org/archives/beast/2017-April/msg00000.html

tim-janik:
Ah, interesting you're posting this here as well. Do you mind filing a pull-request? Just to try out the work flow.

https://help.github.com/articles/about-pull-requests/

tim-janik:
Thanks a lot, pr/6 is merged now.

PR #6, opened 2017-05-13
fix compilation when including bse header for out-of-tree code

swesterfeld:
Here is a pull request which fixes
https://github.com/tim-janik/beast/issues/3

Its the same code I sent to the mailing list before.


tim-janik:
Merged with 4487627828510d6bd8c31c9d1d86724e41ce06ed

PR #14, opened 2017-06-10
Spectmorph plugin

swesterfeld:
Hi!

Here is a branch which moves the SpectMorph plugin from spectmorph repo to beast repo. So while before you had to build beast first, then spectmorph, with this the order is build spectmorph, then beast. This requires the SpectMorph master branch (not any released version), since changes in pkg-config spectmorph*.pc files and installation of additional header was added to allow building the spectmorph beast plugin in beast.

So after the branch has been merged, the build order is

```
# build spectmorph (from spectmorph git master)
./autogen.sh --without-beast            # avoid using libbse and building our own beast plugin
make
make install
# build beast
(as usual)
```

There are a few details you may want to change, namely

1. if the spectmorph package config check fails, this will abort the beast compilation; the alternative would be to output a warning, and to build everything but the spectmorph plugin in this case

2. Qt requires building everything with -fPIC - I added the flag to plugins/spectmorph/gui/Makefile.am, but thats probably not the right place (FIXME is in code)

3. the plugin now still is in the SpectMorph namespace - I don't believe that is much of a problem, but if you feel like changing names and namespaces, please also add compat code so that .bse files which refer to the old SpectMorphOsc still work (I have lots of these files).


tim-janik:
Thanks Stefan, this looks like a really promising start.
Can you please add the following:
a) Make the dependency optional in configure.ac.
b) Adjust the apt-get install list in misc/Dockerfile-apt.in, so "misc/dockerbuild.sh ubuntu:xenial gccbuild" succeeds.
If you need more information on the docker build details (which is what travis-ci checks), let me know and I post something to the mailing list.

swesterfeld:
Should be mergable now. I did some minor cleanups, and addressed your issues:

> a) Make the dependency optional in configure.ac.

I did this now. I also added information at the end of configure which shows which informs the user which optional dependencies will be used. We can add jack support and others there later.

> b) Adjust the apt-get install list in misc/Dockerfile-apt.in, so "misc/dockerbuild.sh ubuntu:xenial gccbuild" succeeds.

Right. At first I tried downloading and installing a binary package (see git log), but I only have one of those for ubuntu-16.04 amd64. Then I used a spectmorph-from-git build, like rapicorn does. However, as from-git spectmorph may be unstable I now use a tag to ensure we get a stable version. So dockerbuild works now.

I also took the opportunity to change the plugin namespace (SpectMorph to Bse::SpectMorph), which I think is appropriate now that it lives inside the beast tree, and added compat code for this (I hope I understood the version here correctly).

swesterfeld:
Ok, now that I released SpectMorph 0.3.4, I also updated the BEAST plugin. Some changes were necessary to support the new SpectMorph 0.3.4 builtin ADSR-Envelope, and since this uses new API, I updated the minimum version you need to compile BEAST to SpectMorph 0.3.4.

The code should be mergeable as-is now.

PR #16, opened 2017-06-28
BEAST: automatically create and connect new bus for new tracks

### LINKED ISSUE #10: Mixer assignment broken
Usually when composing, I want every track on its own mixer channel. There used to be a script (Song -> Assign Tracks to individual Mixer Busses), which is more of a workaround solution. But the script is currently broken. I have no problem with reimplementing the script in Python or whatever language.

Or BEAST could just automatically use one mixer bus per created track. Or it could ask when creating a track: should I create a mixer bus for this track, too. Other DAWs do one of these two options. It could also ask for an instrument to assign to the newly created track. That way track creation would setup the mixer AND set an instrument.

In any case, this is somewhat annoying and could be improved....

swesterfeld:
Fixes https://github.com/tim-janik/beast/issues/10

Signed-off-by: Stefan Westerfeld <stefan@space.twc.de>

### DIFF
--- a/beast-gtk/bsttrackview.cc
+++ b/beast-gtk/bsttrackview.cc
@@ -777,7 +777,9 @@ track_view_action_exec (gpointer data,
 	  track.set_name (string);
 	  g_free (string);
 	  bst_item_view_select (item_view, track.proxy_id());
-          track.ensure_output();
+          Bse::BusH bus = self->song.create_bus();
+          bus.ensure_output();
+          bus.connect_track (track);
 	}
       self->song.ungroup_undo();
       break;
--- a/beast-gtk/bsttrackview.cc
+++ b/beast-gtk/bsttrackview.cc
@@ -787,6 +787,27 @@ track_view_action_exec (gpointer data,
       item = bst_item_view_get_current (item_view);
       track = Bse::TrackH::down_cast (bse_server.from_proxy (item));
       self->song.group_undo ("Delete Track");
+      SfiSeq *seq = nullptr;
+      bse_proxy_get (track.proxy_id(), "outputs", &seq, NULL);
+      BseIt3mSeq *iseq = bse_it3m_seq_from_seq (seq);
+      if (iseq)
+        {
+          for (uint i = 0; i < iseq->n_items; i++)
+            {
+              // automatically remove any output bus which doesn't have other inputs
+              Bse::BusH bus = Bse::BusH::down_cast (bse_server.from_proxy (iseq->items[i]));
+              SfiSeq *bus_inputs_seq = nullptr;
+              bse_proxy_get (bus.proxy_id(), "inputs", &bus_inputs_seq, nullptr);
+              BseIt3mSeq *bus_inputs_iseq = bse_it3m_seq_from_seq (bus_inputs_seq);
+              if (bus_inputs_iseq && bus_inputs_iseq->n_items == 1)
+                {
+                  if (bus != self->song.get_master_bus())
+                    self->song.remove_bus (bus);
+                }
+              bse_it3m_seq_free (bus_inputs_iseq);
+            }
+        }
+      bse_it3m_seq_free (iseq);
       Bse::PartSeq pseq = track.list_parts_uniq();
       self->song.remove_track (track);
       for (const auto &part : pseq)
--- a/bse/bseitem.cc
+++ b/bse/bseitem.cc
@@ -1038,7 +1038,7 @@ ItemImpl::resolve_undo_descriptor_data (const UndoDescriptorData &udd)
   if (udd.upath == "\002project\003")
     return *project;
   BseItem *bitem = bse_container_resolve_upath (bproject, udd.upath.c_str());
-  if (bitem != NULL)    // undo descriptor for NULL objects is not currently supported
+  if (!bitem)    // undo descriptor for NULL objects is not currently supported
     fatal_error ("item undo path failed to resolve");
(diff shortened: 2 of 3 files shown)


tim-janik:
Great, thanks for that work. A few nitpicks, would be nice if you can change the pull request to cater:
git log messages, please use the prefix "BST:" for the beast-gkt dir (and "GXK:" for beast-gtk/gxk). These are originating in the namespace prefix used in those directories.
+      SfiSeq *seq;
+      bse_proxy_get (track.proxy_id(), "outputs", &seq, NULL);
Please per default initialize output variables like "SfiSeq *seq = NULL;", so if we have a bug somewhere, we are more likely to get a NULL pointer crash, rather than random behavior. And about this logic:
+          // automatically remove output bus if
+          //  - we have only one output bus
+          //  - this bus doesn't have any other inputs
Shouldn't we remove all output busses, as long as the track to be deleted is the only input?
That way, no matter what kind of routing is setup, if a user deletes all tracks one by one, there won't be any busses left at the end except for the master bus (assuming the user didn't have any orphan busses to start with).
Oh and, did you test that your patches work across undo/redo (unfortunately we can only test this manually atm, though automated undo testing would be great).

tim-janik:
With your PR applied, the following sequence of actions triggers a warning for me:
  start beast -> Project/New-Song -> Add-Track -> Delete-Mixer/Bus-1 -> Undo -> Undo
Yields:
WARNING: error during undo 'connect_bus' of item "BseBus::Master-1": Parent mismatch

swesterfeld:
> Shouldn't we remove all output busses, as long as the track to be deleted is the only input?

I considered this deletion strategy, and what I don't like about it is that it can delete user-created busses (which were not created by BEAST automatically). For an example assume that we had effects in the mixer.

1. The user creates a track: Snare
2. Then he creates an FX bus, with some effects (Reverb)
3. He connects Snare to FX, to get reverb
4. He creates another Track: HiHats
5. He connects HiHats to FX.

So far, we have two track busses and one FX bus in the mixer

If the user now deletes all tracks (with your deletion strategy), the FX bus will be removed, although it was  manually created. With my deletion strategy, no bus will be removed.

Both are probably suboptimal, but I implemented the "safe" version, which will usually not delete busses it didn't create.

tim-janik:
Ok, I see your reasonings.
I still think we should delete unconnected busses though, because:
a) If the user really intended to keep important stuff around in the bus, there's always Undo if needed.
b) To simplify the user interface in the future, we should probably move towards a 1:1 relationship  between tracks and busses, avoiding orphan busses (or tracks) is a good step into that direction.

swesterfeld:
> Ok, I see your reasonings.
> I still think we should delete unconnected busses though, because:
> a) If the user really intended to keep important stuff around in the bus, there's always Undo if needed.
> b) To simplify the user interface in the future, we should probably move towards a 1:1 relationship between tracks and busses, avoiding orphan busses (or tracks) is a good step into that direction.

Ok, I don't have strong feelings toward either deletion strategy, so I changed it.
I also changed
- BEAST -> BST: in commit messages
- nullptr initialization for SfiSeq*

We still have the undo-problem though:

>With your PR applied, the following sequence of actions triggers a warning for me:
start beast -> Project/New-Song -> Add-Track -> Delete-Mixer/Bus-1 -> Undo -> Undo
> Yields:
> WARNING: error during undo 'connect_bus' of item "BseBus::Master-1": Parent mismatch

I can reproduce this, but I have no idea what would be necessary to fix it (mainly because I'm completely clueless about how undo works internally).

tim-janik:
Hi, please rebase and resolve conflicts here.

> I have no idea what would be necessary to fix it (mainly because I'm completely clueless about how undo works internally).

It's really not that hard to use / model. Every editing API call e.g. add-track() pushes a lambda onto the undo stack that does the exact opposite, i.e. add-track will push remove-track(new-track-id). The only tricky bit here is the object references that have to be passed into the lambdas, because they have to be wrapped to stay persistent. You're best off looking at existing code for that.
Once the user triggers undo, we simply pick the last lambda group from the undo stack (group - since multiple undo lambda calls can comprise a single undo step), and execute it. Since the execution will itself be an API call again, we swap redo/undo stacks around the execution which means that lambda{remove-track()} will actually push another add-track() that lands on the redo stack this way.

swesterfeld:
Ok, I rebased the branch (more like: manually fixed/recommitted what we need). As you say, it is possible to trigger an undo problem like this:

> With your PR applied, the following sequence of actions triggers a warning for me:
> start beast -> Project/New-Song -> Add-Track -> Delete-Mixer/Bus-1 -> Undo -> Undo
> Yields:
> WARNING: error during undo 'connect_bus' of item "BseBus::Master-1": Parent mismatch

However, I investigated the issue and it is not a problem created by the commits in this branch. I've created a new issue for this problem https://github.com/tim-janik/beast/issues/79 and this shouldn't block merging this PR.

tim-janik:
Thank you for your efforts.
Beast development is continued in [Anklang](https://github.com/tim-janik/beast/pull/github.com/tim-janik/anklang/).

For the record, Undo recording is easier in Anklang.
Anklang pushes C++ lambdas onto the undo stack  where objects are pointer-referenced and kept alive via std::shared_ptr<>. 
Undo for objects (track, clip, etc) removal isn't implemented yet, but the plan is to allow repeated unparent/reparent states on objects, which will make Undo a lot easier and avoids dangling references automatically.
The reparenting is something that would have been a lot harder to implement for Beast.

PR #17, opened 2017-06-30
BEAST: fix crash when adding a bus to the mixer while song is playing

swesterfeld:
While the song is playing, adding a bus calls song.create_bus() which returns
NULL, so we cannot call methods on that handle.

Signed-off-by: Stefan Westerfeld <stefan@space.twc.de>


tim-janik:
Thanks a lot, it's merged now.

PR #21, opened 2017-10-09
BLEP based oscillator

swesterfeld:
Here is a plugin that uses BLEPs to implement an oscillator plugin. A few advantages over StandardOsc are:

 - frequency changes over the whole range without aliasing
 - sync without aliasing
 - unison
 - hybrid shapes (between saw and pulse)

I am reasonably sure that this plugin is feature complete at this point. Although some more optimizations may be possible, these will not significantly change the way the plugin sounds (so if they are done, they can be added later on).

(This was initially mentioned here: https://github.com/swesterfeld/beast/issues/9).


swesterfeld:
As for feature complete: I realized that with my implementation so far, sync does affect the main wave form, but not the subharmonic part. Also so far sub is always symmetric, whereas is could have its own pulse width and pulse width modulation.

Right now, I am not sure if we really want to change that, because
- having sync + extra pulse width affect subharmonic component will probably come at a runtime cost
- it is not really easy/straightforward to implement
 
But I thought it would be at least good to mention it here, because if we add it, the sound the bleposc produces for a given parameter set will change (significantly), so it is better to explicitely state this at this point.

swesterfeld:
After my last comment, I started re-implementing the BLEP osc from scratch. Besides all features of my first implementation (unison, hybrid shapes, freq changes/sync without aliasing), I added

- subharmonic oscillator is now affected by sync
- subharmonic oscillator width is configurable via parameter
- minphase blep replaced by linear phase blep (ripples occur before and after jumps)

Merging the plugin into master should make sense now, as I think the plugin code is feature complete at this point. Some performance optimizations may still be possible (for some ideas see TODO), but these can be added later on.

PR #23, opened 2017-10-21
Test out oneline

swesterfeld:
This should be uncontroversial: it makes bsefcompare|testresampler just output one single line:
```
  PASS...
  FAIL...
```
for the tests that are called from make, in order to get less cluttered build output.

If merge are done merging this branch, and also want all commits from resampler-test-fix branch, let me know if I should rebase these for you, as both branches contain changes for testresampler.


tim-janik:
Thanks, merged.

PR #30, opened 2018-02-05
Spectmorph plugin (rebase)

swesterfeld:
Original pull request: https://github.com/tim-janik/beast/pull/14


tim-janik:
Hey Stefan, thanks for resubmitting.
Can you repost the build instructions for spectmorph in this PR? I.e. URL and configure options if any are needed. Also, please submit the configure.ac summary changes seperately, based on what Rapicorn has (showing all dependency libs and paths, etc). We'll merge that first, and the spectmorph and jackd changes can then be based on that.

swesterfeld:
> Can you repost the build instructions for spectmorph in this PR? I.e. URL and configure options if any are needed.

http://spectmorph.org/downloads/spectmorph-0.3.4.tar.bz2

Just the usual: configure, make, make install. The spectmorph plugin will need to be adapted once the new spectmorph ui (without Qt) is ready, but this can take a while. Once spectmorph-0.4.0 with the new ui is out, I'll submit the necessary changes to the beast plugin.

> Also, please submit the configure.ac summary changes seperately, based on what Rapicorn has (showing all dependency libs and paths, etc). We'll merge that first, and the spectmorph and jackd changes can then be based on that.

Yes, will do, this makes sense.

swesterfeld:
> Also, please submit the configure.ac summary changes seperately, based on what Rapicorn has

Configure summary was now merged into master, so I rebased the branch on top of that.

Reaper10:
specmorph icon

swesterfeld:
This plugin uses parts of the SpectMorph API that are no longer supported in the upcoming SpectMorph 0.5.0 and above.

In 0.4.x, the host can
 - manage polyphonic instantiation itself, which this BEAST plugin does, i.e. for a track with 16 voices, there are 16 `MorphPlanVoice` instances, each host voice corresponds to one internal SpectMorph voice - the host interprets note-on/note-off and renders voices
 - leave it to SpectMorph and just send note-on and note-off events using SpectMorph's `MidiSynth`

The instrument editor / WavSource operator introduced in SpectMorph 0.5.0 requires that the voice management is done by SpectMorph, which is what SpectMorph smjack, VST and LV2 have always done. This means that using SpectMorph inside BEAST's modular synthesis engine is no longer possible. With 0.4.x you could add a BEAST filter with BEAST ADSR after the SpectMorph plugin, and it would be done per voice. With 0.5.x this is no longer possible.

So I'd recommend that we don't merge this as-is, because it would allow creating sounds that would no longer be supported by newer SpectMorph (>= 0.5.0). Instead, once we have some kind of Device API, the BEAST-SpectMorph plugin should use it. This would also repair the "Portamento Mono" mode SpectMorph has, which depends on having the midi notes available.

swesterfeld:
I've implemented a preliminary version of a SpectMorph Device for BEAST here: 

  https://github.com/swesterfeld/beast/blob/spectmorph/devices/spectmorph/spectmorph.cc

Some things (custom ui/serialization) are not finished. However, there is a general problem that arises if a user uses multiple versions of SpectMorph at the same time, as the ~/Documents/SpectMorph/Instruments/User directory is shared between all SpectMorph instances. Unexpected results are produced if one version would be linked into BEAST and another version was installed for LV2/VST. Therefore I think to avoid problems with incompatible versions accessing the custom instuments directory, I instead recommend that BEAST uses the LV2 plugin (just like every other DAW). So this PR can be closed now.

PR #31, opened 2018-02-05
Jack driver

swesterfeld:
This branch provides a jack driver for BEAST. The driver will be automatically used if jack is present, otherwise BEAST will fall back to the next best driver. This is audio only, no midi.

See also: discussion on the BEAST list:
 * https://mail.gnome.org/archives/beast/2017-September/msg00003.html


tim-janik:
Ok, thanks, comments:

* configure.ac: work the "Output summary message" into a different patch set, that's to be discussed seperately form a jackd driver. And for what's worth, if we do that, I'd want a format similar to what Rapicorn has, and also add a *lot* more information about other packages there. I.e. that's why it should become a seperate discussion. And, IMHO, software should *never* command a user the way it's done at the end of the summary, that's can easily feel insulting.

* In the comment on "class FrameRingBuffer", add a brief summary in the first line, our doxygen config extracts one-liner briefs from the first line. Same for the other function comments, Add a brief one-liner, before starting with @returns ;-)

* Sigh, emacs suckage, please insert a space between angel brackets when writing "vector<vector<T>>". C++11 can handle it fine but my emacs highlighter still trips up on << >>.

* guint, gfloat, etc. We've stopped using the useless glib type aliases years ago. The only case this actually helps (arguably) is: gchar *string_to_free_via_glib = g_strdup ("..."); So you know to use g_free on the gchar* later on. In any case we definitely don't use these in newly written code, so please adjust by using uint and fix/squash your commit into the first one introducing the types.

* Remove TEST_DROPOUT (unused)

* Move to top of file: define PDEBUG - and, since this is jack, it should be JDEBUG ;-)

* Using '%' is notoriously slow (it's actually a DIV for the CPU), is there a way the ring buffer is likely to be useful with power-of-2 values only? We could use a bit mask instead of '%' then.

* Add the actual values to a warning like this: Bse::warning ("JACK driver: ring buffer size not correct: (jack->buffer_frames != buffer_frames)\n"); I.e. add ": (%u != %u)", so if we get a user report, we can see *what* went wrong, not just where.

* ctors: DeviceDetails() : ports (0), input_ports (0), output_ports (0), etc - Please use direct field initializers in newly written code, i.e. uint ports = 0;

* While you're at it, also always rebase to latest master and do a non-linear push.

* Finally, you have a bunch of trailing whitespaces, please remove. Emacs has a command here, there's probably something in vim as well that can display/kill trailing whitespaces.

swesterfeld:
> configure.ac: work the "Output summary message" into a different patch set, that's to be discussed seperately form a jackd driver. And for what's worth, if we do that, I'd want a format similar to what Rapicorn has, and also add a lot more information about other packages there. I.e. that's why it should become a seperate discussion. And, IMHO, software should never command a user the way it's done at the end of the summary, that's can easily feel insulting.

I've suggested a patch for a configure summary message here: https://github.com/tim-janik/beast/pull/33

Once this is merged, we can rebase this branch.

> In the comment on "class FrameRingBuffer", add a brief summary in the first line, our doxygen config extracts one-liner briefs from the first line. Same for the other function comments, Add a brief one-liner, before starting with @returns ;-)

Done. Although the class is in an anonymous namespace, so doxygen will probably ignore it.

> Sigh, emacs suckage, please insert a space between angel brackets when writing "vector<vector>". C++11 can handle it fine but my emacs highlighter still trips up on << >>.

Done.

> guint, gfloat, etc. We've stopped using the useless glib type aliases years ago. The only case this actually helps (arguably) is: gchar string_to_free_via_glib = g_strdup ("..."); So you know to use g_free on the gchar later on. In any case we definitely don't use these in newly written code, so please adjust by using uint and fix/squash your commit into the first one introducing the types.

We've got quite a bit of history here, so it wasn't easy, but I've eliminated guint and friends.

> Remove TEST_DROPOUT (unused)

I've documented what it does instead, it is used to create dropouts in the driver, from which it should recover properly, I've used this during development.

> Move to top of file: define PDEBUG - and, since this is jack, it should be JDEBUG ;-)

Done.

> Using '%' is notoriously slow (it's actually a DIV for the CPU), is there a way the ring buffer is likely to be useful with power-of-2 values only? We could use a bit mask instead of '%' then.

Well, its not a huge problem, as % it is only called once per block size, so I believe if you were to profile the jack driver, practically no time would be spend here, but hopefully somewhere in the synthesis.

We do have non-power-of-two values, because bse itself has non-power-of-two engine blocks. In fact for jack we could do better on that side, i.e. if the jack callback is 64 values, then our engine block size should probably be 64 values as well. At least I believe that is what everybody else is doing.

However, the ring buffer requires one extra sample space to distinguish between a full and empty buffer. So what we could do is allocate the next power of two as ringbuffer size. As we're only speaking of a speculative optimization that I believe would not make any difference at all here, I'm not particularily keen on changing things, though. It works as it is, is tested, and I don't think it really needs to be changed, or would make a measurable difference anyway.

> Add the actual values to a warning like this: Bse::warning ("JACK driver: ring buffer size not correct: (jack->buffer_frames != buffer_frames)\n"); I.e. add ": (%u != %u)", so if we get a user report, we can see what went wrong, not just where.

Done.

> ctors: DeviceDetails() : ports (0), input_ports (0), output_ports (0), etc - Please use direct field initializers in newly written code, i.e. uint ports = 0;

Done.

> While you're at it, also always rebase to latest master and do a non-linear push.

Done.

> Finally, you have a bunch of trailing whitespaces, please remove. Emacs has a command here, there's probably something in vim as well that can display/kill trailing whitespaces.

Fixed.

swesterfeld:
Ok, as discussed previously, I created a minimal description of how to start jackd and then beast so that we use jackd. I'm not really convinced that this documentation is useful to the average user, but it could help starting jackd if you normally don't use it.

https://github.com/swesterfeld/beast/blob/jack-driver/docs/howto-jack.md

swesterfeld:
> I've suggested a patch for a configure summary message here: #33
> 
> Once this is merged, we can rebase this branch.

Since configure summary was merged, I rebased the jack branch now.

swesterfeld:
I rebased this branch, you should be able to merge this now without problems.

tim-janik:
I've rebased and squashed the PR and then fixed a couple build issues so it actually compiles and works for me (btw namespaces must be labeled with comments in Beast, I had a hard time figuring the beginnins and ends of the anon namespaces). The howto should mention that users also need to install pavucontrol, jackd2 and for building libjack-jackd2-dev. I guess the AppImage needs some extra logic to figure libjack.so is present or not.
The results can be found in this branch: wip/squashed-jack-driver

swesterfeld:
Since you made this a new branch, I'll just add information here instead of committing to my branch.

> The howto should mention that users also need to install pavucontrol, jackd2 and for building libjack-jackd2-dev.

Yes. Either this or the corresponding jack1 packages: jackd1 libjack-dev

In any case I mentioned tentative TODO items for the jack driver during our last phone call, here is a description that can be put as comment into the driver.

```
The jack driver as provided here should be usable. However, there are a few
things that could be improved, here is a short list.

Audio Engine start/stop
-----------------------
Currently the JACK driver registers a new BEAST client every time the device
is opened. This is problematic because connections between the BEAST jack
client and other applications will be disconnected on every close. So
redirecting output song playback through some other application would have
to be reconnected the next time the song plays. Also connecting BEAST to
other JACK clients while the device is closed - before actual playback -
would be impossible.

To fix this, there should be an explicit audio engine start/stop in BEAST.
Once the audio engine is started, the JACK client should remain registered
with JACK, even if no song is playing. This should fix the problems this
driver has due to JACK disconnect on device close.

JACK Midi
---------
Apart from audio, JACK can provide midi events to clients. Note that this
should be optional, and receiving midi events from ALSA should also be
supported.

To receive events from JACK, during setup/open BEAST should register a midi
port at the same place the audio port is registered, so that the BEAST client
will receive both, audio and midi events in the same jack callback.

This breaks with the existing API that audio driver and midi driver should be
implemented seperately, because this requires some data to be shared between
JACK audio / JACK midi driver.

Less buffering, better latency
------------------------------
Currently, the JACK driver has a ring buffer that holds some audio data.  This
introduces latency. This is not what JACK applications typically do.  So here
are some thoughts of how to avoid buffering completely.

To do so, we make the JACK callback block until BEAST has processed the audio
data.

(1) [JACK Thread] jack_process_callback gets called with N input samples
(2) [JACK Thread] wake up engine thread
(3) [JACK Thread] wait for engine thread
(4) [ENGINE Thread] engine thread processes N samples input -> N samples output
(5) [ENGINE Thread] engine thread wakes up jack thread
(6) [JACK Thread] jack_process_callback returns, supplying N output samples

So the idea is to receive input samples from jack (1) and then stall the jack
thread (2). The engine processes the samples (4) and wakes up the jack thread
(5) which then returns (6) the samples the engine created. The engine thread
can use helper threads to complete the job.

Note that this optimization works best (no buffering/latency) if the engine
block size and the jack block size are equal. It also works well if the jack
block size is an integer multiple fo the engine block size.

This avoids latency and buffering. However this may pose stricter RT
requirements onto BEAST. So whether it runs as dropout-free as the current
version would remain to be seen. A not so realtime version would buffer
M complete blocks of N samples, still avoiding partially filled buffers.
```

tim-janik:
I re-read the thread and have a few new comments. But first, can you refresh my mind on whether we discussed for you or me to resolve the conflicts and whether to use the squashed version?

> > Using '%' is notoriously slow (it's actually a DIV for the CPU), is there a
> > way the ring buffer is likely to be useful with power-of-2 values only?
[...]
> As we're only speaking of a speculative optimization that I believe would
> not make any difference at all here, I'm not particularily keen on changing
> things, though. It works as it is, is tested, and I don't think it really
> needs to be changed, or would make a measurable difference anyway.

I brought this up, because I've seen DIV causing heavy performance hits in
otherwise streamlined code several times when profiling and when tracing
slowdowns down to the instruction level. So yes, DIV *does* cause measurable
differences, which is why I default to avoiding it in supposedly fast code.

> I'm not really convinced that this documentation is useful to the average
> user, but it could help starting jackd if you normally don't use it.
> https://github.com/swesterfeld/beast/blob/jack-driver/docs/howto-jack.md

While I don't know *your* average user, there certainly are people/tester/
contributors/etc that don't know the Linux audio stack and history as
intimately as you and can benefit from a description how to get things
to work when using your driver. Please take a look at the Beast manual
(e.g. from the ebeast menu or run `make out/doc/beast-manual.html`) and
make a suggestion where to fit this in. E.g. under '3.2. Howto Jack...' ?

> Apart from audio, JACK can provide midi events to clients. Note that this
> should be optional, and receiving midi events from ALSA should also be
> supported.

Are you saying that when the jack driver is used, Beast should *additionally*
open MIDI devices via ALSA? Do other DAWs do this? At a future point, the
ALSA driver needs to auto-open all MIDI devices anyway so we can start
supporting MIDI auto-discovery. Is that still a good idea when IO should
be routed via Jack though?

Also, please point me at the docs for the exact MIDI event format in Jackd?

[using jack without ring buffer]
> However this may pose stricter RT requirements onto BEAST.

I doubt that, the RT processing of our engine modules must run in bounded
time, which forbids e.g. malloc()/free() calls. Something jack isn't fully
following through in its own implementations (at least last time we discussed
the jack feasibility for Beast). 

I don't think blocking jack for a full buffer calculation in Beast is a good
idea though, one reason to use jack at all is to plug multiple programs
together. What you suggest would severely reduce the CPU resources for other
programs during a PCM block's time frame.

Rather, Beast should be treated like some sort of external synth from jack,
i.e. one jack callback would do:
a) Send new MIDI (audio) data into Beast for the *next* engine block calculation.
b) Fetch the output block of the last engine block calculation to pass on to jack.
c) Wake up the engine so after the callback returns, the engine calculates exactly
   *one* new block that can be fetched upon the next callback invocation.

tim-janik:
An update on how this can be merged. This PR needs:

1) Updates so it uses the new Bse::Driver API and utilizes C++ classes instead of C structures.

2) We need to figure *how* to integrate Jack support into the AppImage. The AppImage build tools gather all system libraries needed by an applicaiton and pack those into the AppImage, so it can be run self-contained. But some libraries are blacklisted and the jack client library is one of those. The reason is that the Jack/libjack ABIs are tightly coupled, so that packing any libjack wouldn't neccessarily work on a system that runs a possibly completely different jackd version.
However, there *are* AppImages out there with Jack support, IIRC they check and compile different libjack versions or something similar. I don't know the exact details, so solving this involves joining #AppImage on freenode, and asking TheAssassin / Probono which AppImages ship jack support so we can investigate those.

swesterfeld:
> 2\. We need to figure _how_ to integrate Jack support into the AppImage.

As you said on IRC, you need one symbol name that was added after all API I require (so if it is present, we can assume libjack is new enough for our driver). I researched this, and here it is:

`jack_get_cycle_times`

* added in 2012 to jack2 (jack2-1.9.9.5) 
* added in 2014 to jack1 (jack1-0.124.0)

So I believe it is highly likely that the distributions our users run ship a version of jack1 or jack2 that is new enough to have `jack_get_cycle_times`.

tim-janik:
Since #128 was merged, this can be closed.

Issue #32, opened 2018-03-07
Crash on part deletion


swesterfeld:
Beast always crashes if I delete a part. To reproduce
 - open party monster
 - delete some part in track view

```
Thread 2 "Beast GUI" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeb953700 (LWP 20887)]
Bse::SongImpl::find_any_track_for_part (this=<optimized out>, part=...) at bsesong.cc:778
778       return track->as<TrackIfaceP> ();
(gdb) bt
#0  Bse::SongImpl::find_any_track_for_part (this=<optimized out>, part=...) at bsesong.cc:778
#1  0x00007ffff7922781 in Bse::__aida_call__Song__find_any_track_for_part (fbr=...) at bseapi_interfaces.cc:6361
#2  0x00007ffff7a2d816 in Aida::ServerConnectionImpl::dispatch (this=0x7fffe409cf20) at ../aidacc/aida.cc:3356
#3  0x00007ffff77b7847 in Bse::AidaGlibSourceImpl::dispatch (this=<optimized out>, this=<optimized out>) at bsestartup.cc:114
#4  Bse::AidaGlibSourceImpl::glib_dispatch (src=<optimized out>) at bsestartup.cc:85
#5  0x00007ffff6120197 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#6  0x00007ffff61203f0 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#7  0x00007ffff612049c in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00007ffff782c7ae in bse_main_loop_thread (init_queue=<optimized out>) at bsemain.cc:222
#9  0x00007ffff5447c80 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00007ffff5ec06ba in start_thread (arg=0x7fffeb953700) at pthread_create.c:333
#11 0x00007ffff4bad41d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:10
(gdb)
```

Possibly related to https://github.com/tim-janik/beast/issues/24

tim-janik:
Thanks for the report, fixed with:

commit ba21e192ce82e143de32e1ff74ac85fb1f76c56b
Date:   Thu Mar 8 01:20:32 2018 +0100
[...]

Issue #24, opened 2017-11-28
Crash on Track Deletion


swesterfeld:
For me, beast (master) crashes if I do
 * New Song
 * Create Track
 * Delete Track

stefan@lemgo:~/src/beast (master % u=)$ gdb beast-0.11.1
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from beast-0.11.1...done.
(gdb) r
Starting program: /usr/local/beast/lib/beast-0-11/bin/beast-0.11.1 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe7efe700 (LWP 9626)]
[New Thread 0x7fffe70d6700 (LWP 9627)]
[New Thread 0x7fffe667f700 (LWP 9628)]
[New Thread 0x7fffe5e7e700 (LWP 9629)]
[New Thread 0x7fffe567d700 (LWP 9630)]
/usr/local/beast/lib/beast-0-11/lib/plugins/spectmorphosc.so: Bse plugin registration failed: Invalid BSE Plugin Version
[New Thread 0x7fffcba1d700 (LWP 9635)]
beast-0.11.1: rcore/aida.cc:1331: const Rapicorn::Aida::RemoteHandle& Rapicorn::Aida::assert_remote_handle(const Rapicorn::Aida::RemoteHandle&): Zusicherung »remote_handle != NULL« nicht erfüllt.

Thread 1 "Beast GUI" received signal SIGABRT, Aborted.
0x00007ffff4234428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
[...]

PR #34, opened 2018-03-08
BSE: bse_cast - an API suggestion of how to handle bse object casts

swesterfeld:
During fixes for https://github.com/tim-janik/beast/issues/32 https://github.com/tim-janik/beast/issues/24 you introduced manual null pointer checks for `as<TrackIfaceP>` and so on. I don't believe that this is the best approach; so I here is a suggestion of how I think it should be done.

I would recommend replacing each invocation of `as<>` with `bse_cast<>`, even if in some cases this will be a little longer:
```
casting track object
old: track ? track->as<TrackIfaceP>() : NULL;
new: bse_cast<TrackIfaceP> (track);

casting this object
old: as<TrackIfaceP>();
new: bse_cast<TrackIfaceP> (this);
```
However, I still think my version is the better API, because
- you cannot forget the null pointer check by accident
- it is a little more intuitive to see that we're just casting here

Its a bit of work to do the details and replace all cases where this is used, but if you want to go this way, I can provide a complete patch which eliminates as<>() completely.

### DIFF
--- a/bse/bseobject.hh
+++ b/bse/bseobject.hh
@@ -57,6 +57,7 @@ typedef enum				/*< skip >*/
 #define BSE_OBJECT_FLAGS_MAX_SHIFT  (16)
 
 /* --- typedefs & structures --- */
+
 struct BseObject : GObject {
   Bse::ObjectImpl       *cxxobject_;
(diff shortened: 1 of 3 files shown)


tim-janik:
Ultimately, the Bse* C-Objects are scheduled for removal now, so I guess this isn't needed anymore.

PR #41, opened 2018-04-21
Hydrogen import

swesterfeld:
Here is my script to import hydrogen drumkits and convert them to .bsewave files. By default it works with the drumkits installed in /usr/share/hydrogen/data/drumkits, but you can also download .debs and import drumkits from a custom directory. I've included instructions in how to download .debs and import from there in the script, just start without any arguments to see instructions.

This branch also includes a small change to libbse so that the bsewave blurb xinfo is displayed at the ui.

### DIFF
--- a/bse/bsewave.cc
+++ b/bse/bsewave.cc
@@ -321,6 +321,9 @@ bse_wave_load_wave_file (BseWave      *self,
                     bse_item_set (self, "uname", wdsc->name, NULL);
                   self->xinfos = bse_xinfos_dup_consolidated (wdsc->xinfos, FALSE);
                   bse_wave_set_locator (self, file_name, wave_name);
+                  const char *blurb = bse_xinfos_get_value (self->xinfos, "blurb");
+                  if (blurb)
+                    bse_item_set (self, "blurb", blurb, NULL);
                 }
               else
                 ; /* error still set from bse_wave_chunk_create() */
--- a/tools/scripts/hydrogen-import.py
+++ b/tools/scripts/hydrogen-import.py
@@ -12,6 +12,8 @@ import sys
 import re
 import subprocess
 import argparse
+import tempfile
+import tarfile
 
 IMPORT_DIR = "/usr/share/hydrogen/data/drumkits"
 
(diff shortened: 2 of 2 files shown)


swesterfeld:
We discussed that beast should be able to download pre-converted drumkits at some point. Here is an example tarball, containing two drumkits (the ones that ship with hydrogen itself).

The download URL is: http://space.twc.de/~stefan/download/beast-drumkits-pack.tar.xz
The tarball contains:

 - Samples/GMkit.bsewave
 - Samples/TR808EmulationKit.bsewave

Let me know if you want me to change

 - the filename
 - the names/paths of the files inside the tarball
 - the format/compression
 - anything else

tim-janik:
@swesterfeld is this something that can be turned into a LiquidSFZ or Anklang bug?

I don't want the work that went into this go to waste, but I wonder what the best way forward is with Anklang taking over from Beast...

swesterfeld:
Basically, liquidsfz tries to transparently load hydrogens xml format, and turns its xml content into .sfz opcodes. For instance if you use
```
$ liquidsfz /usr/share/hydrogen/data/drumkits/GMRockKit/drumkit.xml
```
you can use the hydrogen GMRockKit in liquidsfz. This works because a lot of what the python script has implemented for bsewave files is now also done (even somewhat more correctly as far as I remember) by

* https://github.com/swesterfeld/liquidsfz/blob/master/lib/hydrogenimport.cc

So as soon as Anklang has liquidsfz support, we also have most of what this bug report is about supported in Anklang.

What this doesn't do is
* looking for drumkits in well known folders
* downloading additional h2drumkits
* extracting h2drumkits
* displaying author and/or license infos

(which all could be done by Anklang)

It is also somewhat limited in terms of what hydrogen features are supported, but if we need more, we could improve the liquidsfz importer. In cases where there is no corresponding sfz opcode, we could invent our own private opcodes (i.e. liquidsfz_foo_bar=bazz for a hydrogen foo_bar=bazz feature). In these cases liquidsfz bugs could be filed.

Finally one problem here is that we don't split each drum into its own stereo bus. So you cannot for instance place a reverb on the base drum, but a different effect on the snare, as you only get a mixdown from liquidsfz which contains all drums. This is ok for a start, but suboptimal, but in theory we can implement it by extending the importer and adding new either public or private API to liquidsfz to support splitting its output in more stereo channels. Its a somewhat advanced feature though, and we probably can ignore it for now.

tim-janik:
Great, thanks a lot for this summary!

PR #42, opened 2018-06-03
Fix af-tests/soundfont-test on bionic

swesterfeld:
Soundfont audio tests are currently failing on bionic (18.04). As suggested here:

https://mail.gnome.org/archives/beast/2018-May/msg00001.html

I propose a fix that
- regenerates .ref files to match the 18.04 output
- uses a lower threshold for old fluidsynth versions as to not break the tests on 16.04,...

I was able to find the corresponding fluidsynth changelog entry, which is (fluidsynth 1.1.7):
- fix calculations for modulators (#194)
- https://github.com/FluidSynth/fluidsynth/pull/194

which suggests that the new behaviour is caused by a bugfix.


tim-janik:
Hi Stefan, thanks for the PR.

Overall it looks good, there's one thing that can be improved though.
I'd like to keep all thresholds in one place, in the Makefile.sub. I see two ways to accomplish that:

a) Change configure.ac's FLUID_THRESHOLD to a FLUID_RELAXED_THRESHOLD boolean and pick the threshold inside the Makefile based on the boolean.

b) Change configure.ac's FLUID_THRESHOLD to FLUID_VERSION and pick a threshold inside the Makefile based on FLUID_VERSION>=1.1.7.

I guess (a) suffices for now, but (b) would be more future proof in case we have to adjust the threshold ever again. Come to think of it, we could run into future problems with other libraries as well. Scattering the threshold and version checks across Makefile.sub and configure.ac for each should be avoided. I.e. I'd merge either (a) or (b), but have a slight preference for (b) in terms of setting a precedent for future patches.

swesterfeld:
I see that it would be nicer to have the threshold next to the actual test. I implemented (b), so the version check is now in af-tests/Makefile.sub, using a new helper script based on sort -V - but I doubt we will have this problem often.

tim-janik:
Thanks, merged, plus I re-added your comments.

PR #64, opened 2018-08-23
Port property Song::loop_enabled to C++

swesterfeld:
Please merge. I ported and tested this property port.

Remarks: I tried and failed to use something like (doesn't compile).
```
bool loop_enabled = Bool (... STORAGE SKIP_DEFAULT SKIP_UNDO ...)
```
Everything works as it is, but this would catch typos. In any case, it is not a big problem.

### DIFF
--- a/beast-gtk/bsttrackview.cc
+++ b/beast-gtk/bsttrackview.cc
@@ -81,6 +81,8 @@ bst_track_view_finalize (GObject *object)
     bst_track_roll_controller_unref (self->tctrl);
 
   G_OBJECT_CLASS (bst_track_view_parent_class)->finalize (object);
+
+  delete_inplace (self->song);
 }
 
 GtkWidget*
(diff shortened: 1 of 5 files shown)


tim-janik:
Hi Stefan, looks interesting.

I think I would have kept using iview->container plus a few more casts instead of introducing BstTrackView.song, to ensure that iview->container and BstTrackView.song cannot get out of sync.
Your approach might be more convenient and I guess the bst_item_view_set_container() internals ensure that the two members are always set/reset in sync...

There's one other bit missing though. WIth the old proxy interface, signal connections *always* come in (connect+disconnect) pairs, I'll quote the relevant code lines:

Before your PR:
   bse_proxy_disconnect (song.proxy_id(),
                          "any_signal", track_view_repeat_changed, self,
                          NULL);
  [...]
   bse_proxy_connect (song.proxy_id(),
                         "swapped_signal::property-notify::loop-enabled", track_view_repeat_changed, self,
                         NULL);

After your PR:
      bse_proxy_disconnect (self->song.proxy_id(),
                            "any_signal", track_view_repeat_changed, self,
                            NULL);
      self->song = NULL; /* disconnect */
  [...]
    self->song.on ("notify:loop_enabled", [self]() { track_view_repeat_changed (self); });
    bse_proxy_connect (self->song.proxy_id(),
                         NULL);

I.e. you have replaced proxy_connect() with on(), added an event handler disconnect via =NULL but forgot to remove the proxy_disconnect line for track_view_repeat_changed that belongs to the proxy_connect() you removed.

Also please reword "/* disconnect */" to be more explicit: "// disconnects event handlers"

swesterfeld:
> There's one other bit missing though. WIth the old proxy interface, signal connections always come in (connect+disconnect) pairs
 
> I.e. you have replaced proxy_connect() with on(), added an event handler disconnect via =NULL but forgot to remove the proxy_disconnect line for track_view_repeat_changed that belongs to the proxy_connect() you removed.

Fixed.

> Also please reword "/* disconnect */" to be more explicit: "// disconnects event handlers"

Fixed.

Not sure if you wanted the fixes in an extra commit (I did this) or squashed into the previous commit and delivered as non-linear push.

PR #65, opened 2018-08-23
Port Song::loop_left property

swesterfeld:
One more property C++ified.

### DIFF
--- a/beast-gtk/bsttrackrollctrl.cc
+++ b/beast-gtk/bsttrackrollctrl.cc
@@ -577,7 +577,8 @@ tick_left_move (BstTrackRollController *self,
   if (self->song)
     {
       guint tick = bst_track_roll_controller_quantize (self, drag->current_tick);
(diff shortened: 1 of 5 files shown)


swesterfeld:
After the constant parsing was fixed in master, I rebased and updated this branch to now use SKIP_DEFAULT/SKIP_UNDO.

tim-janik:
Thanks for the PR, a couple comments:
> +      SfiInt lleft = self->song.loop_left();
Use just int, float, etc in new code, instead of gint or SfiInt.

> +  BseSong *self = const_cast<SongImpl*> (this)->as<BseSong *>();
This should be "as<BseSong*>", without the extra space.

> +      bool loop_enabled = self->loop_enabled_SL;
You're saving an old value here for comparing later. There's no need to keep this a variable, i.e. please make that const in similar code in the future:
	+      const bool loop_enabled = self->loop_enabled_SL;

PR #68, opened 2018-08-27
Port Song::loop_right, Song::tick_pointer

swesterfeld:
2 more properties ported to C++

### DIFF
--- a/beast-gtk/bsttrackrollctrl.cc
+++ b/beast-gtk/bsttrackrollctrl.cc
@@ -590,7 +590,8 @@ tick_right_move (BstTrackRollController *self,
   if (self->song)
     {
       guint tick = bst_track_roll_controller_quantize (self, drag->current_tick);
-      bse_proxy_set (self->song, "loop-right", tick, NULL);
+      Bse::SongH song = Bse::SongH::down_cast (bse_server.from_proxy (self->song));
+      song.loop_right (tick);
       drag->state = GXK_DRAG_CONTINUE;
     }
 }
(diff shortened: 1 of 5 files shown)


bse/bsesong.cc
```diff
@@ -991,6 +967,36 @@ SongImpl::loop_right (int tick)
...
+  if (uint (tick) != self->tick_SL)
+    {
+      // this property has no undo
+
+      BSE_SEQUENCER_LOCK ();
+      self->tick_SL = tick;
+      for (SfiRing *ring = self->tracks_SL; ring; ring = sfi_ring_walk (ring, self->tracks_SL))
+        {
+          BseTrack *track = (BseTrack*) ring->data;
+          track->track_done_SL = FALSE;	/* let sequencer recheck if playing */
```
tim-janik: While you're at it, the C macros TRUE and FALSE should be converted to C++'s true and false.

beast-gtk/bsttrackrollctrl.cc
```diff
@@ -565,7 +565,8 @@ pointer_move (BstTrackRollController *self,
       drag->type != GXK_DRAG_DONE) /* skip release events */
     {
       guint tick = bst_track_roll_controller_quantize (self, drag->current_tick);
-      bse_proxy_set (self->song, "tick-pointer", tick, NULL);
+      Bse::SongH song = Bse::SongH::down_cast (bse_server.from_proxy (self->song));
```
tim-janik: It might (soon) be time to convert self->song into a scoped handle. Please keep an eye on when it's easier or not too much effort to use a Bse::SongS in the track controller and go for it, so we can avoid all those from_proxy down casts.

PR #69, opened 2018-08-28
More Song properties C++ified

swesterfeld:
Here is a C++ port for Song: tpqn / numerator / denominator

I renamed tpqn so that it says "TPQN" and not "Ticks" at the UI, because I think "Ticks" is a terrible name for that property, as it doesn't say what kind of ticks we have here.

I added STANDARD_RDONLY - we could also call this STANDARD RDONLY or STANDARD READONLY - I choose the name that SFI_* originally used.

I had to remove the computed defaults, since idl doesn't allow me to compute the default of a property. Previously the default values for properties were set in the Song implementation, so that the property defaults for bpm/numerator/... would always match the defaults from bse_song_timing_get_default().

However I think for Song, duplicating the defaults in bse_song_timing_get_default() and bseapi.idl is okay, because these probably will not change (often). One alternative would be to somehow allow the Song implementation to read the defaults that bseapi.idl defines, and so get rid of the duplication and make the idl file the canonic place.

The only Song property that remains is the post processing network, but since this has the type object, I am not able to port it now. If you have any preference for the next object properties to port, let me know.

### DIFF
--- a/bse/bseapi.idl
+++ b/bse/bseapi.idl
@@ -1013,7 +1013,6 @@ interface Song : SNet {
   // signal void   pointer_changed (int32 a);
   // group _("Timing") {
   // int32   tpqn          = Range  ("Ticks", "Number of ticks per quarter note", STANDARD, 384, 384, 0, 384);
-  // int32   numerator     = Range  ("Numerator", "Measure numerator", STANDARD, 1, 256, 1, 4);
   // int32   denominator   = Range  ("Denominator", "Measure denominator, must be a power of 2", STANDARD, 1, 256, 1, 4);
   // float64 bpm           = Range  ("BPM", "Beats per minute", STANDARD, MIN_BPM, MAX_BPM, 10, 120);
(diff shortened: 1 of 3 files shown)


tim-janik:
Thanks a lot Stefan.
I'm not sure that "TPQN" is any better than "Ticks" for users at the UI, at least I'm not aware of another sequencer that has TPQN labels... I think in either case users will need the tooltip to decipher what the property is about.

There's one other thing I noticed during review. With the old GParamSpec + GObject based property API, the GObject machinery ensures that the GValue passed in to property setters complies with the GParamSpec value range. With our IDL API, nothing like that is enforced anymore. That means, an int32 property between 1-256 can now be set to 0 or -2^31 through the API. For a denominator to become 0 that could affect stability.
Please keep that in mind for ported properties and watch out if we shouldn't add extra guards, like:

  denominator = CLAMP (input, 0, 256)

PR #71, opened 2018-08-28
Remove Rapicorn requirement from README.md

ngeiswei:
As per https://github.com/tim-janik/rapicorn/issues/14#issuecomment-416402264 beast no longer needs rapicorn.


tim-janik:
Good catch, thanks for taking the time to fix the README.

PR #77, opened 2018-09-03
BSE: bseitem: fix undo resolve crash

swesterfeld:
Fix undo for Song::numerator (and others), which causes a crash (fatal error) after clang related updates.


tim-janik:
Thanks a lot Stefan, good catch!
I'd be happy if you reviewed my other changes as well, i.e. what the diffs between branch merges give you:

  git log -p --first-parent

You'd have to skip the odd po/* diff though, not sure if git log can filter that.

PR #78, opened 2018-09-03
Most Bus properties ported to C++

swesterfeld:
I managed to port all Bus properties that have a simple type (non-object, non-sequence) to C++. There are two things worth mentioning here.

(1) in bseapi.idl, we have to use non-computed constants, so there is no elegant way of expressing that the maximum is +24dB - however I think we can live with precomputed factors here

for implementing volume_clamp(), it would be nice if we wouldn't have to duplicate the magic numbers. I originally thought the best way to achieve this would be to be able to access the defaults from the parameter, somewhat like
```
left_volume = CLAMP (val, p_left_volume::min(), p_left_volume::max())
```

However, while implementing the code, I've got a new idea: we could simply offer a way to access the bseapi.idl constants, like
``` 
Const BUS_VOLUME_MIN = 1.58489319246111e-05; /* -96dB */
```
should define a C++ constant, also, and this could be used to implement volume_clamp().

(2) I think C++ properties are commonly referred to as "left_volume" (underscore instead of minus) - I however found the introspected paramspecs in the ui code contain minus, so I use a hacky translation for finding the pspec - not sure if that is the right thing to do here.

### DIFF
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -90,8 +90,25 @@ bus_build_param (BstBusEditor *self,
                  const gchar  *editor,
                  const gchar  *label)
 {
-  GParamSpec *pspec = bse_proxy_get_pspec (self->item, property);
-  self->params = sfi_ring_prepend (self->params, bst_param_new_proxy (pspec, self->item));
+  GxkParam *gxk_param = nullptr;
+
+  /* aida property? */
+  Bse::BusH bus = Bse::BusH::down_cast (bse_server.from_proxy (self->item));
+  for (auto cxxpspec : Bse::introspection_fields_to_param_list (bus.__aida_aux_data__()))
+    if (strcmp (property, cxxpspec->name) == 0)
+      {
+        gxk_param = bst_param_new_property (cxxpspec, bus);
+        break;
+      }
+
+  if (!gxk_param)
+    {
+      /* proxy property */
+      auto pspec = bse_proxy_get_pspec (self->item, property);
+      gxk_param = bst_param_new_proxy (pspec, self->item);
+    }
+
+  self->params = sfi_ring_prepend (self->params, gxk_param);
   GtkWidget *ewidget = gxk_param_create_editor ((GxkParam*) self->params->data, editor);
   gxk_radget_add (self, area, ewidget);
   if (label)
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -83,6 +83,25 @@ bus_editor_release_item (SfiProxy      item,
   bst_bus_editor_set_bus (self, 0);
 }
 
+static GxkParam *
+get_property_param (BstBusEditor *self,
+                    const gchar  *property)
+{
+  Bse::BusH bus = Bse::BusH::down_cast (bse_server.from_proxy (self->item));
+  for (auto cxxpspec : Bse::introspection_fields_to_param_list (bus.__aida_aux_data__()))
+    {
+      std::string pname = cxxpspec->name;
+      for (auto& c : pname)
+        if (c == '-')
+          c = '_';
+
+      if (property == pname)
+        return bst_param_new_property (cxxpspec, bus);
+    }
+
+  return nullptr;
+}
+
 static GtkWidget*
 bus_build_param (BstBusEditor *self,
                  const gchar  *property,
@@ -90,17 +109,7 @@ bus_build_param (BstBusEditor *self,
                  const gchar  *editor,
                  const gchar  *label)
 {
-  GxkParam *gxk_param = nullptr;
-
-  /* aida property? */
-  Bse::BusH bus = Bse::BusH::down_cast (bse_server.from_proxy (self->item));
-  for (auto cxxpspec : Bse::introspection_fields_to_param_list (bus.__aida_aux_data__()))
-    if (strcmp (property, cxxpspec->name) == 0)
-      {
-        gxk_param = bst_param_new_property (cxxpspec, bus);
-        break;
-      }
-
+  GxkParam *gxk_param = get_property_param (self, property); /* aida property? */
   if (!gxk_param)
     {
       /* proxy property */
@@ -152,8 +161,7 @@ bst_bus_editor_set_bus (BstBusEditor *self,
                          "signal::release", bus_editor_release_item, self,
                          NULL);
       /* create and hook up volume params & scopes */
-      pspec = bse_proxy_get_pspec (self->item, "left-volume");
-      GxkParam *lvolume = bst_param_new_proxy (pspec, self->item);
+      GxkParam *lvolume = get_property_param (self, "left_volume");
       GtkWidget *lspinner = gxk_param_create_editor (lvolume, "spinner");
       pspec = bse_proxy_get_pspec (self->item, "right-volume");
       GxkParam *rvolume = bst_param_new_proxy (pspec, self->item);
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -157,19 +157,18 @@ bst_bus_editor_set_bus (BstBusEditor *self,
                          "signal::release", bus_editor_release_item, self,
                          NULL);
       /* create and hook up volume params & scopes */
-      GxkParam *lvolume = get_property_param (self, "left_volume");
-      GtkWidget *lspinner = gxk_param_create_editor (lvolume, "spinner");
-      GxkParam *rvolume = get_property_param (self, "right_volume");
-      GtkWidget *rspinner = gxk_param_create_editor (rvolume, "spinner");
+      GxkParam *volume_db = get_property_param (self, "volume_db");
+      GtkWidget *vspinner = gxk_param_create_editor (volume_db, "spinner");
+      GxkParam *pan = get_property_param (self, "pan");
+      GtkWidget *pspinner = gxk_param_create_editor (pan, "spinner");
+      GtkWidget *panscale = gxk_param_create_editor (pan, "hscale-lin");
+      g_signal_connect_object (panscale, "button-press-event", G_CALLBACK (grab_focus_and_false), pspinner, G_CONNECT_SWAPPED);
       BstDBMeter *dbmeter = (BstDBMeter*) gxk_radget_find (self, "db-meter");
       if (dbmeter)
         {
           GtkRange *range = bst_db_meter_get_scale (dbmeter, 0);
-          bst_db_scale_hook_up_param (range, lvolume);
-          g_signal_connect_object (range, "button-press-event", G_CALLBACK (grab_focus_and_false), lspinner, G_CONNECT_SWAPPED);
-          range = bst_db_meter_get_scale (dbmeter, 1);
-          bst_db_scale_hook_up_param (range, rvolume);
-          g_signal_connect_object (range, "button-press-event", G_CALLBACK (grab_focus_and_false), rspinner, G_CONNECT_SWAPPED);
+          bst_db_scale_hook_up_param (range, volume_db);
+          g_signal_connect_object (range, "button-press-event", G_CALLBACK (grab_focus_and_false), vspinner, G_CONNECT_SWAPPED);
           self->lbeam = bst_db_meter_get_beam (dbmeter, 0);
           if (self->lbeam)
             bst_db_beam_set_value (self->lbeam, -G_MAXDOUBLE);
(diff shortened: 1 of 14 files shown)


tim-janik:
Thanks Stefan, great to see steady progress here.

But note that travis-ci shows, that your changes broke the syndrum audio test in all build variants, please take a look at that.

About volume clamping, taking a look at other DAWs suggests that this should be done at the UI and doesn't need to be based on consts from the IDL at all.
Also the property descriptios read "Volume adjustment in decibel of left/right bus channel" and that's what we should make them. I.e. we should pass actual dB values, e.g. 96 as property value through the IDL layer. (*1)

About GParamSpec containing minus, that's because GLib enforces '-' instead of '_' in the pspec names, so we're forced to work around that in other places as well. Please take a look at name_to_identifier in bstparam.cc, seems like it's time to move that to bstutils.hh to avoid gratituous duplicatoin everywhere.

About porting object / sequence properties, I'm actually not sure what's missing there atm. If you want to make an attempt at porting such a property and nag me about missing bits in the IDL layer or similar, that'd be ok as well.

	+      for (auto& c : pname)

pointers and references are written with *& attached to the variable, i.e. write this as "auto &c"

	+ Const BUS_VOLUME_MIN = 1.58489319246111e-05; /* -96dB */

if at all, this should be using '96' as Const value, and the const be defined in terms of dB. but in other DAWs, the dB range displayed is a function of the UI, e.g. the volume meter of a track display my use different limits than for a related bus channel volume meter.

	+        self->right_volume = self->left_volume = center_volume (self->right_volume, self->left_volume);

While you're at it, please make that a two liner.
Modern code analysis tools tend to choke on using an assignment as rvalue, and its still very readable if it's just:

        self->left_volume = center_volume (self->right_volume, self->left_volume);
        self->right_volume = self->left_volume;


	+      return (self == master);

Please remove extraneous parenthesis like these, they could hide an assignment.
 
	+      auto parent = BSE_ITEM (self)->parent;

Please don't use 'auto' here, but BseItem*, like you did in other places. The reason here is that the variable is the result of a hidden C-style cast (BSE_ITEM()) and used as input for another C-style cast (BSE_SONG), so the actual type used here is completely unclear to the reader and potentially hiding a type bug.

On a side note, it took a while, but I figured again the point of saved_sync. It's clearly trying to preserve the 'sync' state around loading a BSE file (termed 'restore' in the API), and sets that initially to FALSE to allow BSE files to restore different left and right volumes, and syncs thouse only if "sync" is also set or after the restore phase is completed. That heavily depends on the order in which the properties are defined in the BSE file which is somewhat fragile... (*2)

I get the feeling that the old API is a bit too low level with using synced volume factors instead of actual dB values. (*1) and (*2) are both indications that we should probably move towards something like:

* Factors used internally: left_volume, right_volume - though a bit of compat code will be needed to still load these values from old BSE files; syncing is implemented here (as before)
* New API: left_volume_db, right_volume_db - values passed in actual dB, these values can still be different when read even if sync is enabled.
* So essentially, sync, left_volume_db, right_volume_db can all be read/set independently of each other and the internal bus logic adjusts left_volume, right_volume accordingly.

beast-team:
On 09.09.2018 15:21, Stefan Westerfeld via beast wrote:
>   * volume_db = 0, panning = 0 -> left_volume = 1, right_volume = 1
>   * volume_db = 0, panning = -100 -> left_volume = 1.41, right_volume = 0
>   * volume_db = 0, panning = 100 -> left_volume = 0, right_volume = 1.41

What you write here makes sense for libbse. How's the beast-gtk UI going to look
for this though?

> At least using (sqrt(2) ~= 3dB) is what I believe is the correct value here, I'd
> have to double-check that before implementing it.

The monitoring code for our level meters uses dB SPL, which is 20log10(avg) and
corresponds to the power ratio of the signal, so the value to use here would be
more around 2. See also: https://en.wikipedia.org/wiki/Decibel#Acoustics_2

> Also sync is no longer needed
> then, and should be removed.

The sync property is saved im BSE files, so we need compat loading code at
least. If sync is to be removed from the UI depends on how that's designed and
implemented to cope with panning. Thus my question above.

-- 
Yours sincerely,
Tim Janik

https://testbit.eu/timj
Free software author.

swesterfeld:
> On 09.09.2018 15:21, Stefan Westerfeld via beast wrote:
>    * volume_db = 0, panning = 0 -> left_volume = 1, right_volume = 1
>    * volume_db = 0, panning = -100 -> left_volume = 1.41, right_volume = 0
>    * volume_db = 0, panning = 100 -> left_volume = 0, right_volume = 1.41
> What you write here makes sense for libbse. How's the beast-gtk UI going to look
> for this though?

Instead of two volume sliders we only need one. The two meters could be put next to each other. So this would mean left-to-right volume slider (only one slider), scale numbers (-96..+12), left meter, right meter.

Ideal for pan would be a round knob, but we could for now use a slider above the other widgets left to right (where left is -100%, right is +100%).

>  At least using (sqrt(2) ~= 3dB) is what I believe is the correct value here, I'd
>  have to double-check that before implementing it.
> The monitoring code for our level meters uses dB SPL, which is 20log10(avg) and
> corresponds to the power ratio of the signal, so the value to use here would be
> more around 2. See also: https://en.wikipedia.org/wiki/Decibel#Acoustics_2

Lets discuss details not now, but when I've written the code.

>  Also sync is no longer needed
>  then, and should be removed.
> The sync property is saved im BSE files, so we need compat loading code at
> least. If sync is to be removed from the UI depends on how that's designed and
> implemented to cope with panning. Thus my question above.

Right, it should be ignored during read (real compat code is just needed to translate the stored left and right factors => volume/pan). Sync == true files will have identical values for left and right volume, so sync is not needed to load the file, and sync == false files will have their panning information in left and right volume, so this needs to be translated.

There need no longer be a widget for sync either, as using the volume slider affects both, and using pan == 0 is the exact behaviour of our previous sync == true mode.

tim-janik:
> > On 09.09.2018 15:21, Stefan Westerfeld via beast wrote:
> Instead of two volume sliders we only need one. The two meters could be put next to each other. So this would mean left-to-right volume slider (only one slider), scale numbers (-96..+12), left meter, right meter.

Do you intend to work on this as part of this pull request? For beast-gtk and/or ebeast?
I think that's beyond the scope of a property port, and I wonder if we're getting too side tracked here, especially when we're pumping resources into more beast-gtk UI development...

> Ideal for pan would be a round knob, but we could for now use a slider above the other widgets left to right (where left is -100%, right is +100%).
> 
> > At least using (sqrt(2) ~= 3dB) is what I believe is the correct value here, I'd
> > have to double-check that before implementing it.
> > The monitoring code for our level meters uses dB SPL, which is 20log10(avg) and
> > corresponds to the power ratio of the signal, so the value to use here would be
> > more around 2. See also: https://en.wikipedia.org/wiki/Decibel#Acoustics_2
> 
> Lets discuss details not now, but when I've written the code.

When you convert to/from dB, simply use db SPL everywhere (described in the above WP page) then there's no need for further discussions.

> > Also sync is no longer needed
> > then, and should be removed.
> > The sync property is saved im BSE files, so we need compat loading code at
> > least. If sync is to be removed from the UI depends on how that's designed and
> > implemented to cope with panning. Thus my question above.
> 
> Right, it should be ignored during read (real compat code is just needed to translate the stored left and right factors => volume/pan). Sync == true files will have identical values for left and right volume, so sync is not needed to load the file, and sync == false files will have their panning information in left and right volume, so this needs to be translated.

We cannot make any assumptions about the values stored in BSE files, about the order or which values are present or not present, e.g. a user can add sync=1 to a bse file that has left=0.5 and right=1. That's why the code currently ensures left + right are synced after parsing if needed.

FWIW, that's one of the reasons things will get easier when we move to XML based BSE files, we don't have to react to property values coming in in random order, but can query left,right,sync from the XML nodes and attributes and configure the object accordingly.

swesterfeld:
Ok, so let me comment on anything that doesn't have to do with the new volume_db/pan properties first.

> Thanks Stefan, great to see steady progress here.
> 
> But note that travis-ci shows, that your changes broke the syndrum audio test in all build variants, please take a look at that.

Right. This is a really nasty problem, which is caused by loading/reading code for float64 properties in aida. It turns out that even simple floating point numbers like "1.03" are written out correctly, but restore breaks in bsestorage.cc, because the scanner_parse_paren_rest does return this as "1. 3" and this is restored as a value of 1.0. I've fixed this by using the existing sfi_scanner_parse_real_num function to parse float64 values (see also commit comments). This seems to work, tests pass again. Not sure if this is what you think we should do to fix the problem, though.

> About volume clamping, taking a look at other DAWs suggests that this should be done at the UI and doesn't need to be based on consts from the IDL at all.
> Also the property descriptios read "Volume adjustment in decibel of left/right bus channel" and that's what we should make them. I.e. we should pass actual dB values, e.g. 96 as property value through the IDL layer. (*1)

Ok. You'll notice that the magic factors are gone, and the clamping looks sane (CLAMP (v, -96, 12)). In particular I don't try to enforce restrictive CLAMP() boundaries for left_factor and right_factor, they are now an implementation detail and should not be used by the UI.

> About GParamSpec containing minus, that's because GLib enforces '-' instead of '_' in the pspec names, so we're forced to work around that in other places as well. Please take a look at name_to_identifier in bstparam.cc, seems like it's time to move that to bstutils.hh to avoid gratituous duplicatoin everywhere.

Fixed.

> About porting object / sequence properties, I'm actually not sure what's missing there atm. If you want to make an attempt at porting such a property and nag me about missing bits in the IDL layer or similar, that'd be ok as well.

Ok. Maybe I'll try it later.

> ```
> +      for (auto& c : pname)
> ```
> pointers and references are written with *& attached to the variable, i.e. write this as "auto &c"

Ok.

> ```
> + Const BUS_VOLUME_MIN = 1.58489319246111e-05; /* -96dB */
> ```
> if at all, this should be using '96' as Const value, and the const be defined in terms of dB. but in other DAWs, the dB range displayed is a function of the UI, e.g. the volume meter of a track display my use different limits than for a related bus channel volume meter.

Fixed.

> ```
> +        self->right_volume = self->left_volume = center_volume (self->right_volume, self->left_volume);
> ```
> While you're at it, please make that a two liner.

Fixed. But code was deleted in a later commit.

> +      return (self == master);
> ```
> Please remove extraneous parenthesis like these, they could hide an assignment.

Fixed.

> ```
> +      auto parent = BSE_ITEM (self)->parent;
> ```
> Please don't use 'auto' here, but BseItem*, like you did in other places. The reason here is that the variable is the result of a hidden C-style cast (BSE_ITEM()) and used as input for another C-style cast (BSE_SONG), so the actual type used here is completely unclear to the reader and potentially hiding a type bug.

Fixed.

> On a side note, it took a while, but I figured again the point of saved_sync. It's clearly trying to preserve the 'sync' state around loading a BSE file (termed 'restore' in the API), and sets that initially to FALSE to allow BSE files to restore different left and right volumes, and syncs thouse only if "sync" is also set or after the restore phase is completed. That heavily depends on the order in which the properties are defined in the BSE file which is somewhat fragile... (*2)

Right. Consider this fixed, as we no longer need sync, and I see absolutely no point in trying to be too perfect here. We read left_volume and right_volume in any order, and that is it. The UI should always use the volume_db/pan properties. If the user used sync before saving, left_volume and right_volume will be the same, and pan will be 0. If not, left_volume and right_volume will be different, and pan will not be 0.


So after commenting on your remarks, let be briefly describe what I did to get volume_db/pan properties. First, I looked at some variations on how to compute factors like our left_volume/right_volume from a total volume_db and panning. A good description of the possibilities is here:

http://www.cs.cmu.edu/~music/icm-online/readings/panlaws/index.html

I choose to implement constant power panning (-3dB), because:
- some popular DAWs use it as default (for instance Cubase)
- the mathematical definition is quite straightforward (we want constant power)
- from this follows that we have all functions we need to implement translation backwards/forwards

However this seems to be one of the things that don't have a globally optimal solution. For instance I tried Bitwig, expecting to see constant power panning, but they appear to use -4.5dB pan law (also confirmed on Forums). Cubase afaik has a project wide setting for -6dB, -4.5dB, -3dB, 0dB, constant power (default).

Long story short: I implemented the volume_db/pan properties as a view on the underlying left_volume/right_volume factors with constant power panning (-3dB). Constant power panning should be a good solution, and it gives the user the ability to control volume and panning individually. We can change the default later on, if we choose to implement other panning strategies.

I adapted the beast-gtk UI. As we have no knob widget, I use a horizontal slider to represent panning. I believe that the new UI is definitely an improvement over the old from a usability point of view, even if it may take some initial effort to understand which widget is doing what.


You'll also find one somewhat clumsy commit that allows setting a Bus::master_output property with and without undo. The old framework had the possibility to suppress undo via g_object_set. I have here implemented one method with and one without undo. However a framework solution would probably be better in the long run, such as
{
  auto no_undo = bus->scoped_undo_blocker();
  bus->master_output (true); // not undoable
}

The other case would be
{
  bus->master_output (true); // undoable
}

tim-janik:
So, to summarize our last meeting:
- The bus properties should be ported as straight forward as possible, without adding new functionality / UI.
- The APPLY_IDL_PROPERTY() macro should be used.

Once you've done this, please re-assign the new code to me.

swesterfeld:
> So, to summarize our last meeting:
> 
> * The bus properties should be ported as straight forward as possible, without adding new functionality / UI.
> * The APPLY_IDL_PROPERTY() macro should be used.
> 
> Once you've done this, please re-assign the new code to me.

I'm doing this in an new branch/PR as to keep the old version as reference: so the new code is now found in https://github.com/tim-janik/beast/pull/105

tim-janik:
> I'm doing this in an new branch/PR as to keep the old version as reference: so the new code is now found in #105

Thanks, we're continuing the development in #105 then.

PR #101, opened 2019-04-12
Fluidsynth2

swesterfeld:
This updates the fluidsynth code to fluidsynth >= 2.0.0. This is not backward compatible, BEAST will not build against fluidsynth 1 any longer. The soundfont audio test reference file had to be updated due to a  bugfix in fluidsynth 1 related to voice panning.


swesterfeld:
One remark about us using the deprecated fluid_nwrite_float() function: if I use the non-deprecated version, our soundfont audio unit test fails. And things start sounding somewhat crappy. Turns out that the new replacement function is buggy, and we trigger this. I submitted a fix to upstream.

https://github.com/FluidSynth/fluidsynth/pull/527

But since it will take time until this propagates into official fluidsynth releases I recommend staying with the old deprecated API for some time.

PR #102, opened 2019-04-23
One fluid synth per track

swesterfeld:
This gets rid of all bse module locks in the fluid synth code, by using one fluid synth instance per track.

I observed that if we load the soundfont every time the user presses play, this can be slow (50ms to load fluid r2 * number of soundfont tracks), so this keeps the fluid synth instances even after the project stops playing, and re-uses the instances on play.

What this doesn't implement is changing the soundfont on a track while the project is playing (the gui doesn't let me do this anyway), but this could be implemented by preparing a new fluid_synth instance and exchanging it on the fly using an engine job.

I believe that this is the best/correct way to do it, better than https://github.com/tim-janik/beast/pull/85

### DIFF
--- a/bse/bsesoundfont.cc
+++ b/bse/bsesoundfont.cc
@@ -12,8 +12,6 @@
 
 #define parse_or_return         bse_storage_scanner_parse_or_return
 
-using std::string;
-
 enum {
   PARAM_0,
   PARAM_FILE_NAME,
@@ -186,7 +184,7 @@ bse_sound_font_reload (BseSoundFont *sound_font)
   return bse_sound_font_load_blob (sound_font, sound_font_impl->blob, FALSE);
 }
 
-string
+std::string
 bse_sound_font_get_filename (BseSoundFont *sound_font)
 {
   Bse::SoundFontImpl *sound_font_impl = sound_font->as<Bse::SoundFontImpl *>();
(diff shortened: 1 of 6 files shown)


swesterfeld:
As requested, I've updated this branch so that it doesn't require fluidsynth >= 2.0.5.

bse/bsesoundfontrepo.cc
```diff
@@ -387,10 +277,6 @@ SoundFontRepoImpl::~SoundFontRepoImpl ()
       delete_fluid_settings (fluid_settings);
       fluid_settings = NULL;
```
tim-janik: It'd be better to merge the above two lines into `delete_fluid_settings (&fluid_settings);` where `delete_fluid_settings` automatically assigns `NULL` to the pointer passed in.

swesterfeld: `delete_fluid_settings` is part of the fluid synth api, so I cannot change what it does

bse/bsesoundfont.cc
```diff
@@ -12,6 +12,8 @@
 
 #define parse_or_return         bse_storage_scanner_parse_or_return
 
+using std::string;
```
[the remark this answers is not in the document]
swesterfeld: Right, I've fixed this now by using std::string - I also tried using Bse::String - this doesn't have any advantage here, as all code is outside the Bse namespace.

swesterfeld:
> The reason the GUI blocks instrument changes is that the BSE core hasn't had support for this yet. If soundfonts can implement changes during runtime, the associated, we'll of course release that restriction and make it user accessible. I guess that's better left for another PR, right?
 
I agree. As I understand it, updating beast-gtk is to be avoided, I suppose this should be done once ebeast supports assigning soundfont & preset with the new ui. The fluid code as it is supports changing the preset of an existing track during play. It doesn't support changing the SF2 filename of a track during play, but this could be implemented by swapping the fluid_synth_t instance in this case. Maybe that should be done after you have the new engine module api.

Anyway I think I've addressed the other issues now, will reassign this to you so you can merge it.

PR #105, opened 2019-05-05
Bus properties ported to C++ - redone with APPLY_IDL_PROPERTY() macro

swesterfeld:
As requested, I went through all changes in https://github.com/tim-janik/beast/pull/78 and used APPLY_IDL_PROPERTY() where possible. In some cases it wasn't possible, so there are a few handwritten setters. I also avoid any fancy new features in the code  or ui, this is an 1:1 port.

This also fixes the C++ property loader code in bsestorage.cc so that make check passes.

From my perspective the only thing that could be controversial is that you used g_object_set(...property...value...) to avoid undo recording for some setter calls. I added a function to block undo manually so

```
  g_object_set (self, /* no undo */
               "sync", self->saved_sync,
                NULL);
```

now looks somewhat like this

```
  bus->block_property_undo(); 
  bus->sync (self->saved_sync);
  bus->unblock_property_undo();
```

This should work in the same way for setters that use APPLY_IDL_PROPERTY() and for hand-written setters that use push_property_undo(). If you have a better suggestion feel free to either change this after merging or letting me know how it should be done.

### DIFF
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -89,8 +89,24 @@ bus_build_param (BstBusEditor *self,
                  const gchar  *editor,
                  const gchar  *label)
 {
-  GParamSpec *pspec = bse_proxy_get_pspec (self->item, property);
-  self->params = sfi_ring_prepend (self->params, bst_param_new_proxy (pspec, self->item));
+  GxkParam *gxk_param = nullptr;
+
+  /* aida property? */
+  Bse::BusH bus = Bse::BusH::__cast__ (bse_server.from_proxy (self->item));
+  const Bse::StringSeq meta = bus.find_prop (property);
+
+  GParamSpec *cxxpspec = Bse::pspec_from_key_value_list (property, meta);
+  if (cxxpspec)
+    gxk_param = bst_param_new_property (cxxpspec, bus);
+
+  if (!gxk_param)
+    {
+      /* proxy property */
+      auto pspec = bse_proxy_get_pspec (self->item, property);
+      gxk_param = bst_param_new_proxy (pspec, self->item);
+    }
+
+  self->params = sfi_ring_prepend (self->params, gxk_param);
   GtkWidget *ewidget = gxk_param_create_editor ((GxkParam*) self->params->data, editor);
   gxk_radget_add (self, area, ewidget);
   if (label)
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -82,6 +82,20 @@ bus_editor_release_item (SfiProxy      item,
   bst_bus_editor_set_bus (self, 0);
 }
 
+static GxkParam *
+get_property_param (BstBusEditor *self,
+                    const gchar  *property)
+{
+  Bse::BusH bus = Bse::BusH::__cast__ (bse_server.from_proxy (self->item));
+  const Bse::StringSeq kvlist = bus.find_prop (property);
+
+  GParamSpec *cxxpspec = Bse::pspec_from_key_value_list (property, kvlist);
+  if (cxxpspec)
+    return bst_param_new_property (cxxpspec, bus);
+
+  return nullptr;
+}
+
 static GtkWidget*
 bus_build_param (BstBusEditor *self,
                  const gchar  *property,
@@ -89,15 +103,7 @@ bus_build_param (BstBusEditor *self,
                  const gchar  *editor,
                  const gchar  *label)
 {
-  GxkParam *gxk_param = nullptr;
-
-  /* aida property? */
-  Bse::BusH bus = Bse::BusH::__cast__ (bse_server.from_proxy (self->item));
-  const Bse::StringSeq meta = bus.find_prop (property);
-
-  GParamSpec *cxxpspec = Bse::pspec_from_key_value_list (property, meta);
-  if (cxxpspec)
-    gxk_param = bst_param_new_property (cxxpspec, bus);
+  GxkParam *gxk_param = get_property_param (self, property); /* aida property? */
 
   if (!gxk_param)
     {
@@ -150,8 +156,7 @@ bst_bus_editor_set_bus (BstBusEditor *self,
                          "signal::release", bus_editor_release_item, self,
                          NULL);
       /* create and hook up volume params & scopes */
-      pspec = bse_proxy_get_pspec (self->item, "left-volume");
-      GxkParam *lvolume = bst_param_new_proxy (pspec, self->item);
+      GxkParam *lvolume = get_property_param (self, "left_volume");
       GtkWidget *lspinner = gxk_param_create_editor (lvolume, "spinner");
       pspec = bse_proxy_get_pspec (self->item, "right-volume");
       GxkParam *rvolume = bst_param_new_proxy (pspec, self->item);
--- a/beast-gtk/bstbuseditor.cc
+++ b/beast-gtk/bstbuseditor.cc
@@ -84,35 +84,36 @@ bus_editor_release_item (SfiProxy      item,
 
 static GxkParam *
 get_property_param (BstBusEditor *self,
+                    Bse::BusH     bus,
                     const gchar  *property)
 {
-  Bse::BusH bus = Bse::BusH::__cast__ (bse_server.from_proxy (self->item));
   const Bse::StringSeq kvlist = bus.find_prop (property);
 
-  GParamSpec *cxxpspec = Bse::pspec_from_key_value_list (property, kvlist);
-  if (cxxpspec)
-    return bst_param_new_property (cxxpspec, bus);
-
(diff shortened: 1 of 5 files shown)


bse/bsebus.cc
```diff
@@ -1050,6 +1011,55 @@ BusImpl::right_volume (double val)
...
+}
+
+void
+BusImpl::master_output (bool val)
+{
+  BseBus *self = as<BseBus*>();
+
+  if (val != master_output())
+    {
+      auto prop = "master_output";
```
tim-janik: Add `const`

bse/bsestorage.cc
```diff
@@ -741,6 +741,16 @@ storage_parse_property_value (BseStorage *self, const std::string &name, Bse::An
 {
   assert_return (BSE_IS_STORAGE (self), G_TOKEN_ERROR);
   GScanner *scanner = bse_storage_get_scanner (self);
+  if (any.kind() == Aida::FLOAT64) /* float format cannot be handled by scanner_parse_paren_rest */
```
tim-janik: I don't think any.kind() will always hold the desired type here for all code paths.
This is more a workaround than a fix, I think scanner_parse_paren_rest should be fixed instead.

bse/bsebus.cc
```diff
@@ -1014,13 +1014,13 @@ BusImpl::right_volume (double val)
...
-  BseBus *self = const_cast<BusImpl*> (this)->as<BseBus*>();
-  auto parent = BSE_ITEM (self)->parent;
+  BseBus  *self = const_cast<BusImpl*> (this)->as<BseBus*>();
+  BseItem *parent = BSE_ITEM (self)->parent;
   if (BSE_IS_SONG (parent))
     {
       BseSong *song = BSE_SONG (parent);
       BseBus *master = bse_song_find_master (song);
-      return (self == master);
+      return self == master;
```
tim-janik: Code like this can be merged down into the culprit of your current branch. Here's how you do that:

1) Use `git blame HEAD -- bsebus.cc` to find the faulty commit:
```
aaabbbccc   bse/bsebus.c swesterfeld     return (self == master);
```
2) Commit the fix as in:
```
git commit -m '+++ fix commit aaabbbccc' 
111222333
```
3) Assuming your current branch started from master, merge this down with: `git rebase -i master`
4) Move `pick 111222333 +++ fix commit aaabbbccc` into a `squash` or `fixup` following the culprit:
```
pick aaabbbccc BSE: introduce buggy code
squash 111222333 +++ fix commit aaabbbccc
```

bse/bsesong.cc
```diff
@@ -497,8 +497,10 @@ bse_song_ensure_master (BseSong *self)
     {
       BseUndoStack *ustack = bse_item_undo_open (self, "Create Master");
       child = (BseSource*) bse_container_new_child_bname (BSE_CONTAINER (self), BSE_TYPE_BUS, master_bus_name(), NULL);
-      g_object_set (child, "master-output", TRUE, NULL); // not-undoable
       Bse::BusImpl *bus = child->as<Bse::BusImpl*>();
+      bus->block_property_undo();
```
tim-janik: Get rid of the block/unblock here, I'd like this to be handled elsewhere.

bse/bsesong.cc
```diff
@@ -681,7 +683,7 @@ SongImpl::remove_bus (BusIface &bus_iface)
   BseItem *child = bus.as<BseItem*>();
   BseUndoStack *ustack = bse_item_undo_open (self, __func__);
   // reset ::master-output property undoable
```
tim-janik: The comment about undoable here can be removed.

bse/bseitem.cc
```diff
@@ -1335,4 +1337,16 @@ ItemImpl::apply_idl_property_need_undo (const StringVector &kvlist)
   return !Aida::aux_vector_check_options (kvlist, "", "hints", "skip-undo");
 }
 
+void
+ItemImpl::block_property_undo()
+{
```
tim-janik: I really don't want to introduce these. We already have other mechanism that check if undo steps should be recorded at all, and we have a dummy undo stack `bse_undo_stack_dummy` that could be used if we'd want to temporarily ignore undo steps. Also, the *C++* developer shouldn't be bothered with having to use block/unblock in pairs, so for the future, similar API should make use of ctor/dtor pairs like std::lock_guard does.

Assuming your branch starts off from `master`, you can get rid of this commit with `git rebase -i` and then remove the `pick ...` line that refers to this commit.

beast-gtk/bstbuseditor.cc
```diff
@@ -89,8 +89,24 @@ bus_build_param (BstBusEditor *self,
                  const gchar  *editor,
                  const gchar  *label)
 {
-  GParamSpec *pspec = bse_proxy_get_pspec (self->item, property);
-  self->params = sfi_ring_prepend (self->params, bst_param_new_proxy (pspec, self->item));
+  GxkParam *gxk_param = nullptr;
```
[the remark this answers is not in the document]
swesterfeld: Right. This has been refactored in a later commit, so the issue is effectively resolved.

bse/bsebus.cc
```diff
@@ -1003,6 +993,25 @@ BusImpl::BusImpl (BseObject *bobj) :
...
+  return self->muted;
+}
+
+void
+BusImpl::mute (bool val)
+{
+  BseBus *self = const_cast<BusImpl*> (this)->as<BseBus*>();
+
+  if (APPLY_IDL_PROPERTY (self->muted, val))
+    {
```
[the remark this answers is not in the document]
swesterfeld: Ok, the lines caused by { } are not really necessary, so I removed then. I'd still like to keep a blank line between BseBus *self and the rest of the code. Obviously "readability" is not an objective criterium but depends on the coder reading the code. But for me, the extra blank line makes the code more readable. I often use blank lines to seperate code in groups that belong together. And we're no longer coding with 80x24 terminals, so we have space for many lines on the screen.

beast-gtk/bstbuseditor.cc
```diff
@@ -89,8 +89,24 @@ bus_build_param (BstBusEditor *self,
...
+
+  /* aida property? */
+  Bse::BusH bus = Bse::BusH::__cast__ (bse_server.from_proxy (self->item));
```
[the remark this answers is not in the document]
swesterfeld: I fixed this in a later commit.

beast-gtk/bstbuseditor.cc
```diff
@@ -89,8 +89,24 @@ bus_build_param (BstBusEditor *self,
...
+  const Bse::StringSeq meta = bus.find_prop (property);
```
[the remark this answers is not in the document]
swesterfeld: Fixed in a later commit.

bse/bsebus.cc
```diff
@@ -1004,14 +981,52 @@ BusImpl::mute() const
...
+}
+
+void
+BusImpl::solo (bool is_solo)
+{
+  BseBus *self = as<BseBus*>();
+
+  if (solo() != is_solo)
+    {
+      auto prop = "solo";
```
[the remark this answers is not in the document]
swesterfeld: Fixed.

bse/bsestorage.cc
swesterfeld: Right I removed the code now, so make check fails, so this needs to be fixed (probably by you) now by fixing the parser.

bse/bsebus.cc
swesterfeld: Thanks for the hint, I normally don't need this while coding, but I tried to fix everything here that could be done by history rewriting using git rebase -i now.

bse/bsesong.cc
swesterfeld: Ok, I removed the block/unblock stuff now.

bse/bsesong.cc
swesterfeld: Fixed.

bse/bseitem.cc
swesterfeld: Ok, I just tried to do a 1:1 port so I had to do something where the old code blocked undo. If you say you believe it is not necessary, thats even better, because the code looks cleaner now. I removed block/unblock now.

swesterfeld:
> Please fix review comments I left for you and remove:
> a) the commit introducing the block/unblock API,

Done with all code that depends on it.

> b) the float64 workaround.

Done.

> If that means you updated PR is broken, point out the bugs you're runing into and I'll give you a hand with the fixes. In particualy, I'm not even sure you can trigger buggy behaviour due to not blocking undo, if you find a way to trigger that, that'd be really helpful.

I haven't ever seen any undo related problem while testing it.

> As for fixing the float parsing, I looked into that earlier and am quite confident I can fix parse_paren_rest. I just needed an actual test case that triggered breakage, which I didn't have at the time.

Right. `make check` is broken now due to audio tests, so before merge, you should look into fixing the parser.

PR #109, opened 2019-07-28
Track primitive properties ported to C++

swesterfeld:
I've ported all Track properties to C++ that do not contain object references. Two things are worth mentioning.

1. `Const MIDI_MAX_CHANNEL` - I declared this as constant, but maybe the definition must be moved elsewhere. It seems to be required in bsebasics.idl, too (there is a `// FIXME: MAX_MIDI_CHANNEL` in the `MidiChannelEvent` definition)
2. I didn't touch `"signal::property-notify::n-voices"` and similar signal code in bsttrackview.cc, I'm not sure what the C++ property equivalent would be

### DIFF
--- a/beast-gtk/bsttrackview.hh
+++ b/beast-gtk/bsttrackview.hh
@@ -20,11 +20,13 @@ typedef	struct	_BstTrackView	   BstTrackView;
 typedef	struct	_BstTrackViewClass BstTrackViewClass;
 struct _BstTrackView
 {
-  BstItemView	          parent_object;
-  Bse::SongS              song;
-  BstTrackRoll	         *troll;
-  BstTrackRollController *tctrl;
-  GtkWidget		 *repeat_toggle;
+  BstItemView                     parent_object;
+  Bse::SongS                      song;
+  std::map<SfiProxy, Bse::TrackS> track_map;
+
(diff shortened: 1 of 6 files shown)


swesterfeld:
> I didn't touch `"signal::property-notify::n-voices"` and similar signal code

**Update:** I ported this to C++ now using `track.on("notify:n_voices",...)` for the new properties.

beast-gtk/bsttrackview.hh
```diff
@@ -20,11 +20,13 @@ typedef	struct	_BstTrackView	   BstTrackView;
...
 struct _BstTrackView
 {
-  BstItemView	          parent_object;
-  Bse::SongS              song;
-  BstTrackRoll	         *troll;
-  BstTrackRollController *tctrl;
-  GtkWidget		 *repeat_toggle;
+  BstItemView                     parent_object;
+  Bse::SongS                      song;
+  std::map<SfiProxy, Bse::TrackS> track_map;
```
tim-janik: Note, I'm perfectly happy with adding type aliases for complex types, so you don't need to reformat all members. E.g.

    using TrackMap = std::map<SfiProxy, Bse::TrackS>;
    // other members...
    TrackMap    track_map;

PR #110, opened 2019-07-28
Port Super author/license to C++

swesterfeld:
This is the first string property port that uses the `APPLY_IDL_PROPERTY` macro. Since there is no version of `constrain_idl_property` for strings, your code didn't compile when instantiated with strings. So I used C++17 `if constexpr` for checking for the type which compiles now. This means strings are not constrained (and all other types). If that wasn't your intention when writing the code, probably a String specialization of `constrain_idl_property` would be better.

I preserved the quark setting/getting (no idea why it is done this way in the first place). However I wonder if SuperImpl shouldn't simply have two string members, one for author and one for license.

### DIFF
--- a/bse/bseapi.idl
+++ b/bse/bseapi.idl
@@ -878,6 +878,8 @@ interface Container : Source {
 
 /// Base interface type for Item managers.
 interface Super : Container {
+  String author  = String (_("Author"), _("Person changing or creating this object"), STANDARD);
+  String license = String (_("License"), _("Copyright license applying to this object"), STANDARD);
 };
 
 /// A list of Super type objects.
--- a/bse/bseitem.hh
+++ b/bse/bseitem.hh
@@ -235,7 +235,7 @@ public:
       return false;
     T rvalue = cvalue;
     const StringVector kvlist = find_prop (propname);
-    if (std::is_integral<T>::value || std::is_floating_point<T>::value || std::is_enum<T>::value)
+    if constexpr (std::is_integral<T>::value || std::is_floating_point<T>::value || std::is_enum<T>::value)
       {
         if (!constrain_idl_property (rvalue, kvlist))
           return false;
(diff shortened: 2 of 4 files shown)


swesterfeld:
I also don't know how to port sfi_pspec_time typed properties. And copyright looks like it can be removed (?), has a `// COMPAT-FIXME: remove around 0.7.0`

swesterfeld:
As discussed,
* I removed quark based properties and replaced them with String members
* I removed the copyright property (which only very old files would have)

swesterfeld:
This triggers a problem in your `scanner_parse_paren_rest` modifications to bsestorage.cc (also appears to break travis CI):
```
stefan@quadcorn:~/src/ghbeast (pport-super *% u=)$ make -j1 check
[...]
  CHECK    out/tests/bsefiles.lst-b
bsetool: bse/bsestorage.cc:726: scanner_parse_paren_rest: assertion failed: text1 > text0 && text1[-1] == ')'
./media/Demos/x2-midi-test.bse:106: error: failure around character ')' - aborting...
check-load: ./media/Demos/x2-midi-test.bse: loading failed: Bse.Error.PARSE_ERROR
tests/Makefile.mk:84: recipe for target 'out/tests/bsefiles.lst-b-test' failed
make: *** [out/tests/bsefiles.lst-b-test] Error 1
```

From what I could debug, this is a problem of your code accessing the internal GScanner buffer (`scanner->text`). It looks like the buffer was refilled during the loop that should read the property, so that text1 is smaller than text0. One way to fix it could be to load the entire file into memory before parsing.

tim-janik:
> From what I could debug, this is a problem of your code accessing the internal GScanner buffer (`scanner->text`). It looks like the buffer was refilled during the loop that should read the property, so that text1 is smaller than text0. One way to fix it could be to load the entire file into memory before parsing.

Thanks a lot for an accurate analysis and suggesting a proper fix, great input!

PR #112, opened 2019-08-03
Property ports for MidiSynth

swesterfeld:
This ports all properties of the MidiSynth to C++ (except for object properties SNet/PNet).

I had to fix a bug in BsePcmOutput (first commit), because setting the volume on the output object had no effect, but I wanted to test my code.

To have 1:1 the same ranges, I introduced constants for BSE_MIN|MAX_VOLUME_dB, but maybe this should just be hardcoded to -96 / 24. For the new max midi channel constant I think a constant should be used (maybe moved up to the other constants).

I noticed that some property idl constants don't combine properly. If you write Range (... GUI STORAGE ...), this expands to "r:w:G" "r:w:S" - so instead of getting a "G" hint, you get a "Gr" hint. Here I used STANDARD as ported property type, to work around the issue,which has both "G" and "S" hints. Not sure what is intended here, but adding ":" at the beginning and end of each property constant would allow using both versions for a GUI & STORAGE property.

### DIFF
--- a/bse/bseapi.idl
+++ b/bse/bseapi.idl
@@ -1138,8 +1138,9 @@ interface MidiSynth : SNet {
                                 1, 256, 1, 16);
   };
   group _("Adjustments") {
-    float64 volume_f  = Range (_("Master [float]"), "", STORAGE, 0, 15.8489319246111 /* +24dB */, 0.1, 1.0);
-    float64 volume_dB = Range (_("Master [dB]"), "", GUI ":dial", MIN_VOLUME_DB, MAX_VOLUME_DB, 0.1, 0);
+    float64 volume_f    = Range (_("Master [float]"), "", STORAGE, 0, 15.8489319246111 /* +24dB */, 0.1, 1.0);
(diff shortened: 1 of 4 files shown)


bse/bseapi.idl
```diff
@@ -1135,6 +1135,11 @@ interface MidiSynth : SNet {
                                 STANDARD ":scale",
                                 1, 256, 1, 16);
   };
+  group _("Adjustments") {
+    float64 volume_f = Range (_("Master [float]"), _("Master volume as factor"),
+                              STORAGE,
+                              0, 15.8489319246111 /* +24dB */, 0.1, 1.0);
```
tim-janik: 15.84... used to be calculated as a factor from MAX_dB, if we every adjust MAX-dB in the IDL file, this needs to be adjusted as well, so it shouldn't be hardcoded.
I.e. just add another constant next to the MAX_dB constant definition and use that for this registration, so it's obvious that both have to be adapted if we touch the constants in the future.

PR #115, opened 2019-08-08
Item properties ported to C++

swesterfeld:
Actually this is just one property: `seqid`

Since no more properties remain, I wonder if I should add a commit which removes the setter/getter (in this case `bse_item_set|get_property_internal`).


tim-janik:
> Actually this is just one property: `seqid`
> 
> Since no more properties remain, I wonder if I should add a commit which removes the setter/getter (in this case `bse_item_set|get_property_internal`).

Haven't reviewed yet, but yes, that sounds like a good plan.

PR #120, opened 2019-08-28
BSE: randomhash.hh: fix compilation on g++ 7.4.0

swesterfeld:
std::string in C++17 cannot be used at compile time, so constexpr is not possible


tim-janik:
fnv1a_consthash64() is named consthash because it proivdes a compile-time hash function, removing the constexpr means it cannot be used at compile time. It's a pitty that g++ doesn't handle this as well as clang, I'm removing the entire function for now.

PR #122, opened 2019-09-03
Ladder Filter

swesterfeld:
This is the implementation of a Moog-style lowpass ladderfilter with saturation and 2x oversampling. The number of poles is configurable . Filter modes are available for 6db/octave, 12db/octave, 18db/octave and 24db/octave.

I don't have any TODOs left right now, so this could either be the final version, or at least fairly close to the final version. I'm discarding the commit history here, this is not so interesting. The necessary changes for the resampler (https://github.com/tim-janik/beast/pull/121) are included in this branch.

Right now the resampling is done with the FPU (as plugins/bseblockutils.cc is not getting compiled), but that is still fast enough for daily use. I've also noticed that BSE doesn't enforce SSE-alignment of input/output buffers, so right now SSE support is broken; if that were fixed, performance would be a bit better.

Of course this depends a bit on which inputs/outputs are connected, but for the common case one (stereo) ladder filter takes
 * 0.4% CPU usage with FPU resampling
 * 0.33% CPU usage with SSE resampling


tim-janik:
It'd be great to have this ported to the new AudioSignal::Processor API and also be able to enable key-tracking

Issue #119, opened 2019-08-09
No sound without -p alsa=pulse


swesterfeld:
BEAST for a very long time now refused to produce audio out of the box without the command line args `-p alsa=pulse` - I am using  (setup by Cadence)
 - JACK as primary sound server
 - PulseAudio bridged to JACK

Problems have been reported by other users (https://github.com/tim-janik/beast/issues/92 https://github.com/tim-janik/beast/issues/80). Since the new Electron UI doesn't have command line args, the workaround cannot be used there, so the new UI has no sound.

In order to make this work out-of-the-box: why does the workaround work, whereas auto-detected audio output fails?

First of all, it doesn't matter if you use `-p alsa=default` or `-p alsa=pulse`. The default device is pulse. So this is the first device the ALSA driver tries to open during auto detect. But open fails during auto detect. The reason here is that during auto detect, BEAST tries to open the `default` device in full duplex mode. Why doesn't this work?

Using printf()s in the alsa pcm driver, I found that although `read_handle` and `write_handle` can be opened for `pulse`, the following fails
```
   if (error == 0 && alsa->read_handle && alsa->write_handle &&
       snd_pcm_link (alsa->read_handle, alsa->write_handle) < 0)
     error = Bse::Error::DEVICES_MISMATCH;
```
This was observed already here: https://github.com/tim-janik/beast/issues/80#issuecomment-421115474

In any case, the auto detection will try all available devices in duplex mode, and settle for some audio device that I don't use - instead of the default device - like for instance onboard audio which is not connected.

If I on the other hand specify `-p alsa=pulse`, BEAST will try to open the audio output in duplex mode, and fail. Then it will try to open pulseaudio in output-only mode, and succeed.

Some ideas how to make it work out-of-the-box:
- adapt alsa driver so that it still works if `snd_pcm_link` fails
- merge JACK driver
- allow explicit driver selection in the UI (effectively allowing users to force PulseAudio output)
- force default device if default device is present, even if this has no full duplex mode

tim-janik:
Beast development is continued in [Anklang](github.com/tim-janik/anklang/) which has Jack support.
If you think snd_pcm_link errors should be ignored during auto detection, please make a case there.

PR #123, opened 2019-09-04
Force alsa default device

swesterfeld:
This is a fix for #119 - and finally should make beast work out of the box for users which have pulseaudio as default device (for instance together with jack/bridging). For me at least this fixes "make run".

In order to force using the default device if available, pcm drivers can return a string which device is the default device.

### DIFF
--- a/bse/bsedevice.cc
+++ b/bse/bsedevice.cc
@@ -88,6 +88,15 @@ bse_device_open (BseDevice      *self,
   assert_return (BSE_IS_DEVICE (self), Bse::Error::INTERNAL);
   assert_return (!BSE_DEVICE_OPEN (self), Bse::Error::INTERNAL);
   Bse::Error error = Bse::Error::DEVICE_NOT_AVAILABLE;
+  if (!arg_string)
+    {
+      /* if the user didn't specify a device, we use the default device of the
+       * driver (if the driver has a default device) - for ALSA this typically
+       * means we will use "default" as arg_string
+       */
+      if (BSE_DEVICE_GET_CLASS (self)->default_device)
+        arg_string = BSE_DEVICE_GET_CLASS (self)->default_device (self);
+    }
   if (arg_string)
     error = device_open_args (self, need_readable, need_writable, arg_string);
   else
--- a/bse/bsedevice.hh
+++ b/bse/bsedevice.hh
@@ -51,6 +51,7 @@ struct BseDeviceClass : BseObjectClass {
   void                (*post_open)     (BseDevice    *device);
   void                (*pre_close)     (BseDevice    *device);
   void                (*close)         (BseDevice    *device);
+  const char*         (*default_device) (BseDevice   *device);
 };
 struct BseDeviceEntry {
   BseDevice      *device;
--- a/drivers/bsepcmdevice-alsa.cc
+++ b/drivers/bsepcmdevice-alsa.cc
@@ -189,6 +189,25 @@ bse_pcm_device_alsa_list_devices (BseDevice *device)
   return ring;
 }
 
+static const char *
+bse_pcm_device_alsa_default_device (BseDevice *device)
+{
+  /* we check if the "default" device is available */
+  SfiRing *entries = nullptr;
+  entries = list_pcm_devices (device, entries, "Duplex");
+  entries = list_pcm_devices (device, entries, "Output");
+
+  const char *default_device = nullptr;
+  for (SfiRing *node = entries; node; node = sfi_ring_walk (node, entries))
+    {
+      BseDeviceEntry *entry = (BseDeviceEntry*) node->data;
+      if (strcmp (entry->device_args, "default") == 0)
+        default_device = "default";
+    }
+  bse_device_entry_list_free (entries);
+  return default_device;
+}
+
 static void
 silent_error_handler (const char *file,
                       int         line,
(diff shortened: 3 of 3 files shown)


tim-janik:
This patch causes a regression on my system (Ubuntu 18.04 with pulseaudio) when opening media/Demos/stereo-through.bse. Without the patch, I can play this:
    
    BSE_DEBUG=snd pasuspender -- out/beast-gtk/beast-0.15.0 out/media/Demos/stereo-through.bse 
    snd: bse_device_open: BsePcmDeviceALSA: default: Bse.Error.DEVICES_MISMATCH
    snd: bse_device_open: BsePcmDeviceALSA: null: Bse.Error.DEVICE_BUFFER
    snd: bse_device_open: BsePcmDeviceALSA: pulse: Bse.Error.DEVICES_MISMATCH
    snd: bse_device_open: BsePcmDeviceALSA: sysdefault:CARD=PCH: Bse.Error.DEVICES_MISMATCH
    snd: bse_device_open: BsePcmDeviceALSA: front:CARD=PCH,DEV=0: Bse.Error.NONE

FYI, without pasuspend, opening 'front' gives Bse.Error.FILE_BUSY.
So fixing 'default' as the only ALSA device to open doesn't work for opening in Duplex mode with pasuspend. Note that even without pulseaudio ([killing pulseaudio](https://askubuntu.com/questions/8425/how-to-temporarily-disable-pulseaudio/394872#394872)) running, I'm getting:

       >        default:CARD=PCH - HDA Intel PCH, ALC298 Analog Default Audio Device
       >        front:CARD=PCH,DEV=0 - HDA Intel PCH, ALC298 Analog Front speakers
    snd: bse_device_open: BsePcmDeviceALSA: default:CARD=PCH: Bse.Error.DEVICES_MISMATCH
    snd: bse_device_open: BsePcmDeviceALSA: front:CARD=PCH,DEV=0: Bse.Error.NONE

So I'm not sure *what* ALSA picks as 'default' (without pulse), but it doesn't have matching mixfreq/fragment sizes which are needed for duplex mode.
(BSE_DEBUG=snd is supported by recent master)

tim-janik:
A bit of evaluation on Ubuntu 18.04:

- Bitwig defaults to opening JACK and issues a warning if that fails. It provides JACK, ALSA and PulseAudio drivers. For the ALSA driver, the user has to select one of the available ALSA hardware drivers (hw:*). 
- Reaper defaults to opening JACK and issues a warning if that fails. It does allow the user to manually select ALSA, a dummy driver or PulseAudio. Oddly enough, with ALSA it only recognizes the first hardware device. Reaper allows the project to render with a different sample rate than used by the audio device.
- Renoise tries to open ALSA's *first* hw device (hw:0,0), its ALSA driver allows manual selection of all ALSA hardware devices (hw:*). If ALSA hw:0 doesn't work, it tries to fallback to its second driver JACK, if that doesn't work it gives up.
- Firefox and Google-chrome simply use Pulseaudio out of the box and support Duplex operation. Via Pulseaudio, it's possible to change between Headphone and Laptop mic/speakers while a call is ongoing. Since they can also mix-in remote audio, they must have some kind of drift-compensation.

Insisting on using a JACK driver when Jackd is not running or not even installed on a system is simply user-unfriendly. Renoise knows better and can play out of the box via ALSA, but if that doesn't work, it's worse of than any of the other programs, because it fails to recognize/use PulseAudio.

The browsers are a magnitude more user friendly than any of the DAWs, they support PulseAudio out of the box, try to minimize latency for interactive calls, work in *Duplex mode out of the box* and compensate drift automatically.

For Beast, we should simply work as well and as user-friendly as the browsers by picking PuleAudio by default, trying JACK only if Jackd is running and falling back to the first ALSA hw device if it is not busy. That way, power users can manually assign JACK as the default driver and for everyone else, things will work out of the box. But we need some kind of drift compensation when we use PulseAudio in Duplex mode.

tim-janik:
> Ok, first of all, I no longer think that changing the auto-detect order for devices can solve our problem.

I think we can get very close to 0-config for most users by doing this:

1) Pick Jack driver if and only if Jackd is running.
2) Pick Pulseaudio if it is available, note that this does not support snd_pcm_link and needs some drift compensation.
3) Pick the first ALSA HW device, that should support snd_pcm_link.
4) Otherwise leave it to the user via the preferences dialog.

The only thing I'm not perfectly sure about is if a linked HW device (3) is better than Pulseaudio with drift compensation (2). From a quality standpoint, (3) would be clearly better.
But imagine a user has Pulseaudio hooked up to his *second* ALSA device, e.g. USB headphones. If Beast prefers the first ALSA device over Pulseaudio, it can link Duplex channels, but may blast onto laptop speakers unintentionally or maybe utterly silent if hw:0,0 is a desktops onboard sound with an unconnected stereo jack.

> So we need explicit driver selection in the future. So eventually we will need .bserc to have an entry for setting the audio driver. I can understand that adapting beast-gtk may not be what we want to do, since it makes not much sense to invest time in legacy ui code.

I've almost thrown away the entire preferences code in beast-gtk yesterday, because it depends on the Visitor code which doesn't compile due to a g++-8 bug. I did find a workaround in the end though. But don't count on that code to stay around much longer.

> I can also understand that adapting ebeast may not yet be what we want to do, because it may need more time to develop ebeast until it is time to add a configuration dialog. But that will happen eventually.

Ebeast has had a preferences dialog since mid-May.

We just need to add PCM device selection to it.

tim-janik:
Ebeast has PCM / MIDI device selection now.

PR #124, opened 2019-09-10
BSE: bsemathsignal: add approximations: Bse::fast_log2 and Bse::fast_exp2

swesterfeld:
First of all, this adds a function that was missing in bsemathsignal: a `fast_log2` implementation. It works by using the float exponent and approximating the rest of the value using a polynomial. I found a tool that gives optimal polynomial coefficients for a given order: https://github.com/samhocevar/lolremez

Since I was at it already, I also looked at our exp2 approximations. Comparing them to what lolremez would produce, what we do is not ideal. So I also added `fast_exp2`. This has consistently better relative errors. So I suggest as next step replacing `bse_approxN_exp2` by `fast_exp2<N>` everywhere, which should be as fast, but more accurate. Please review the API and let me know if this can be the canonical API for porting the old functions to.

Remarks:
* unlike `bse_approxN_exp2`, our new `fast_exp2<N>` no longer guarantees almost zero error if the argument is an integer; however I think we don't really need to preserve this property 
* I like `Bse::fast_log2` better than `Bse::approx_log2` because when reading code "fast" gives me an idea what the function does and why it was used instead of log2

Relative `exp2(x)` approximation errors in interval [-1:1]:

```
rxprec: bse_approx2_exp2: 0.009017387
rxprec: bse_approx3_exp2: 0.0007944462
rxprec: bse_approx4_exp2: 5.568432e-05
rxprec: bse_approx5_exp2: 3.24224e-06
rxprec: bse_approx6_exp2: 1.614916e-07
rxprec: bse_approx7_exp2: 7.028907e-09
rxprec: bse_approx8_exp2: 2.716875e-10
rxprec: bse_approx9_exp2: 9.445048e-12

rxprec: fast_exp2<2, double>: 0.001724763
rxprec: fast_exp2<3, double>: 7.478144e-05
rxprec: fast_exp2<4, double>: 2.593371e-06
rxprec: fast_exp2<5, double>: 7.493647e-08
rxprec: fast_exp2<6, double>: 1.8558e-09
rxprec: fast_exp2<7, double>: 4.021119e-11
rxprec: fast_exp2<8, double>: 7.746751e-13
rxprec: fast_exp2<9, double>: 1.375958e-14
```

### DIFF
--- a/bse/bsemathsignal.hh
+++ b/bse/bsemathsignal.hh
@@ -706,4 +706,275 @@ bse_cent_tune_fast (int fine_tune)
   return bse_cent_table[CLAMP (fine_tune, -100, 100)];
 }
 
+namespace Bse {
+
+/**
+ * @param  value positive input value
+ * @return y     approximating log2 (value)
+ *
+ * Template arguments:
+ *   ORDER  tradeoff between speed and accuracy of the approximation
+ *   T      whether to use double or float for internal calculations
+ *
+ * This function computes a fast approximation for log2. The first template
+ * argument ORDER must be specified. The second selects the floating point
+ * type and defaults to double.
+ *
+ * fast_log2<2> (value)
+ *   is the fastest version with the largest approximation error
+ * fast_log2<9> (value)
+ *   is the slowest version with the smallest approximation error
+ *
+ * In cases where only a float value is required, the second template argument
+ * T can be set to float. Typically this will be a bit faster than using double
+ * precision internally. In that case the ORDER can be at most 6, since for
+ * higher orders the error introduced by float computations is too high.
+ *
+ * Maximum approximation error in interval [1;2] for different values ORDER is
+ * (first value T = double, second value T = float)
+ *
+ *   2  0.00493976    0.004940044
+ *   3  0.0006371173  0.0006375115
+ *   4  8.759192e-05  8.858764e-05
+ *   5  1.253874e-05  1.412891e-05
+ *   6  1.845687e-06  4.548069e-06
+ *   7  2.772895e-07  -
+ *   8  4.231443e-08  -
+ *   9  6.53724e-09   -
+ *
+ * The error is computed as: | log2 (value) - fast_log2 (value) |
+ */
+template<int ORDER, typename T = double>
+static inline T G_GNUC_CONST
+fast_log2 (float value)
+{
+  union {
+    float f;
+    int i;
+  } float_u;
+  float_u.f = value;
+  // compute log_2 using float exponent
+  const int log_2 = ((float_u.i >> 23) & 255) - 128;
+  // replace float exponent
+  float_u.i &= ~(255 << 23);
+  float_u.i += BSE_FLOAT_BIAS << 23;
+
+  static_assert (ORDER >= 2 && ORDER <= 9);
+  if constexpr (std::is_same<float, T>::value)
+    static_assert (ORDER <= 6);
+
+  T x = float_u.f;
+
+  // polynomials designed with: lolremez -d <ORDER> -r 1:2 "log(x)/log(2)+1"
+  if constexpr (ORDER == 2)
+    {
+      T u = -3.4484843300001946e-1;
+      u = u * x + T (2.0246657790474698);
+      x = u * x + T (-6.748775860711561e-1);
+    }
+  else if constexpr (ORDER == 3)
+    {
+      T u = 1.5824870260836496e-1;
+      u = u * x + T (-1.0518750217176431);
+      u = u * x + T (3.0478841468943746);
+      x = u * x + T (-1.1536207104929913);
+    }
+  else if constexpr (ORDER == 4)
+    {
+      T u = -8.1615808498122383e-2;
+      u = u * x + T (6.4514236358772082e-1);
+      u = u * x + T (-2.1206751311142674);
+      u = u * x + T (4.0700907918522014);
+      x = u * x + T (-1.5128546239033371);
+    }
+  else if constexpr (ORDER == 5)
+    {
+      T u = 4.4873610194131727e-2;
+      u = u * x + T (-4.1656368651734915e-1);
+      u = u * x + T (1.6311487636297217);
+      u = u * x + T (-3.5507929249026341);
+      u = u * x + T (5.0917108110420042);
+      x = u * x + T (-1.8003640347009253);
+    }
+  else if constexpr (ORDER == 6)
+    {
+      T u = -2.5691088815846394e-2;
+      u = u * x + T (2.7514877034856807e-1);
+      u = u * x + T (-1.2669182593669425);
+      u = u * x + T (3.2865287704176774);
+      u = u * x + T (-5.3419892025067624);
+      u = u * x + T (6.1129631283200212);
+      x = u * x + T (-2.0400402727100282);
+    }
+  else if constexpr (ORDER == 7)
+    {
+      T u = 1.5125359168008401e-2;
+      u = u * x + T (-1.8393964232233078e-1);
+      u = u * x + T (9.7809484881021081e-1);
+      u = u * x + T (-2.9850211023586282);
+      u = u * x + T (5.7814391203775074);
+      u = u * x + T (-7.494130176731051);
+      u = u * x + T (7.1339697617832616);
+      x = u * x + T (-2.2455378914374758);
+    }
+  else if constexpr (ORDER == 8)
+    {
+      T u = -9.0889081045466651e-3;
+      u = u * x + T (1.2384570293556399e-1);
+      u = u * x + T (-7.4835742226622081e-1);
+      u = u * x + T (2.6387738689074552);
+      u = u * x + T (-6.0134326106109641);
+      u = u * x + T (9.2859836735642522);
+      u = u * x + T (-1.0007135250742891e+1);
+      u = u * x + T (8.1548040722981051);
+      x = u * x + T (-2.4253930836663341);
+    }
+  else if constexpr (ORDER == 9)
+    {
+      T u = 5.547594786690081e-3;
+      u = u * x + T (-8.3767270477714422e-2);
+      u = u * x + T (5.6751026819556446e-1);
+      u = u * x + T (-2.2749487308928086);
+      u = u * x + T (5.9910685362749332);
+      u = u * x + T (-1.0884909361115433e+1);
+      u = u * x + T (1.3970237262129601e+1);
+      u = u * x + T (-1.2880952951531813e+1);
+      u = u * x + T (9.1755128331255977);
+      x = u * x + T (-2.585298173957386);
+    }
+
+   return x + log_2;
+}
+
+/**
+ * @param  ex exponent within [-127..+127]
+ * @return y  approximating 2^ex
+ *
+ * Template arguments:
+ *   ORDER  tradeoff between speed and accuracy of the approximation
+ *   T      whether to use double or float for internal calculations
+ *
+ * This function computes a fast approximation for exp2. The first template
+ * argument ORDER must be specified. The second selects the floating point
+ * type and defaults to double.
+ *
+ * fast_exp2<2> (ex)
+ *   is the fastest version with the largest approximation error
+ * fast_exp2<9> (ex)
+ *   is the slowest version with the smallest approximation error
+ *
+ * In cases where only a float value is required, the second template argument
+ * T can be set to float. Typically this will be a bit faster than using double
+ * precision internally. In that case the ORDER can be at most 6, since for
+ * higher orders the error introduced by float computations is too high.
+ *
+ * Maximum relative approximation error in interval [-1;1] for different
+ * values ORDER is (first value T = double, second value T = float)
+ *
+ *   2  0.001724763   0.001724847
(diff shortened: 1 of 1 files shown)


tim-janik:
> Relative `exp2(x)` approximation errors in interval [-1:1]:

That's a miniscule range to approximate, given the valid range of this function. Here are the errors of 
bse_approx3_exp2() at mostly integer steps with some omissions for brevity:

       -2.0, +0.00000000000000000
       -1.5, -0.00016133294256113
       -1.0, +0.00000000000000000
       -0.5, -0.00032266588512226
       +0.0, +0.00000000000000000
       +0.5, -0.00112351661969536
       +1.0, +0.00000000000000000
       +1.5, -0.00224703323939071
       +2.0, +0.00000000000000000
       +2.5, -0.00449406647878142
       +7.0, +0.00000000000000000
      +11.0, +0.00000000000000000
      +16.0, +0.00000000000000000
      +32.0, +0.00000000000000000
      +40.0, +0.00000000000000000
      +48.0, +0.00000000000000000
      +54.0, +0.00000000000000000
      +64.0, +0.00000000000000000
     +127.0, +0.00000000000000000

Now the same for fast_exp2<2>():

       -2.0, +0.00011078548906696
       -1.5, -0.00060979588254370
       -1.0, +0.00022157097813391
       -0.5, -0.00121959176508741
       +0.0, +0.00044314195626782
       +0.5, +0.00243918353017435
       +1.0, +0.00088628391253565
       +1.5, +0.00487836706034869
       +2.0, +0.00177256782507129
       +2.5, +0.00975673412069738
       +7.0, +0.05672217040228134
      +11.0, +0.90755472643650137
      +16.0, +29.04175124596804380
      +32.0, +1903280.20965576171875000
      +40.0, +487239733.671875
      +48.0, +124733371820
      +54.0, +7982935796480
      +64.0, +8174526255595520
     +127.0, +75396696880394894980022595129180160

I.e. a pure Remez approximation works allmost as well as approxX_exp2 while using one less addmul, but only within -1:+1. I've discarded that approach when I wrote approxX_exp2, because outside of that range the error becomes gigantic. Above, I'm comparing approx3_exp2 with fast_exp2<2>, to show that with just one extra addmul, those giant errors can be avoided (which even fast_exp2<9> cannot accomplish) if the integer part and the fractional part are approximated separately.
Considering that fast_exp2<>() are not even minimizing the error at exactly -1, +1, or at 0 at the very least, I'd at most add them as rough approximations to be used only when the input range is known to be -1..+1 (which is not the case for general audio signals in BSE). For that it might be named fast_narrow_exp2<>() or similar, but I think fast_exp2<>() is better used if we rename the current bse_approxX_exp2 functions.

tim-janik:
> First of all, this adds a function that was missing in bsemathsignal: a `fast_log2` implementation. It works by using the float exponent and approximating the rest of the value using a polynomial. I found a tool that gives optimal polynomial coefficients for a given order: https://github.com/samhocevar/lolremez

Note that your code adds a needless cast that costs time and precision, the formulas all contain terms like:

      u = u * x + T (constant); // T ∈ { float, double, long double }

For e.g. T=float, this costs precision. X86 FPUs load the constants into one of the internal FPU registers, the internal FPU registers are 80bit wide, casting the constant to float before hand truncates some of the last digits without making the following FPU internal operations any faster.
A similar precision loss can be observed with AMD64, example:

    fast_log2<6, double> (+1.5)
        0.00000054181458465
    fast_log2<6, float> (+1.5)
        0.00000153622037214

I.e. I'd strongly recommend to remove that cast.
Also, Wikipedia suggests that similarly to exp2(), a log2 approximation can be split into a quick integer part approximation and a severely constrained approximation for the fractional part that only affects the intervall [1,2): https://en.wikipedia.org/wiki/Binary_logarithm#Iterative_approximation

So I'd be interested to see the Remez approximation optimized for only [1,2) (with an error of 0 at [1], this can be achieved by subtracting a constant). So combined with adding the integer part, the approximated function still matches log2() at integer points exactly.

tim-janik:
As mentioned on IRC, enabling optimizations with MODE=release and picking clang++ (6.0 here) vs g++ (7.4 here) makes major differences when benchmarking exp2f and log2f from glibc against our approximations. On a modern AMD64 processor, glibc is often faster. Internally it also uses polynomials around order 4, but picks its coefficients from a table depending on the input argument. With that it achieves errors <  1 ULP and is often speedier because it can also use hand crafted SSE2 implementations.
I haven't had a chance to benchmark the approximations on musl, but so far, based on your submission, I'm inclined to integrate the following:

1) Rename bse_approx6_exp2 to fast_exp2() and get rid of the other approximation variants.
2) Add fast_log2() based on your 6th order version, but with error correction for integer logarithms.
3) When building for AMD64, use exp2f to implement fast_exp2 and use log2f to implement fast_log2.

Here's the error correction I'm talking about, note that exchanging "long double" for "float" makes the code significantly slower, because it forces the compiler to add code to reduce precision. On my machine, this version is roughly as fast as log2f when compiling with optimizations, with both compilers:

    static inline long double G_GNUC_CONST
    fast_log2f (float value)
    {
      union {
        float f;
        int i;
      } float_u;
      float_u.f = value;
      // compute log_2 using float exponent
      const int log_2 = ((float_u.i >> 23) & 255) - 128;
      // replace float exponent
      float_u.i &= ~(255 << 23);
      float_u.i += BSE_FLOAT_BIAS << 23;
      long double u, x = float_u.f;
      // lolremez --long-double -d 6 -r 1:2 "log(x)/log(2)+1-0.00000184568668708"
      u =         -2.5691088815846393966e-2l;
      u = u * x +  2.7514877034856806734e-1l;
      u = u * x + -1.2669182593669424748l;
      u = u * x +  3.2865287704176774059l;
      u = u * x + -5.3419892025067624343l;
      u = u * x +  6.1129631283200211528l;
      x = u * x + -2.040042118396715321l;
      return x + log_2;
    }

Error samples, compared to LOG2L(3):

       +0.0, -0.00000231613294631
       +0.5, +0.00000000000000000
       +1.0, +0.00000000000000000
       +1.1, -0.00000181973000285
       +1.5, -0.00000130387210186
       +1.8, -0.00000312228549678
       +2.0, +0.00000000000000000
       +2.2, -0.00000181973000285
       +2.5, -0.00000140048214306
       +3.0, -0.00000130387210186
       +4.0, +0.00000000000000000
       +5.0, -0.00000140048214306
       +6.0, -0.00000130387210186
       +7.0, -0.00000312228549678
       +8.0, +0.00000000000000000
       +9.0, -0.00000084878575295
      +10.0, -0.00000140048214306
      +11.0, -0.00000368176020430
      +16.0, +0.00000000000000000
      +32.0, +0.00000000000000000
      +40.0, -0.00000140048214306
      +48.0, -0.00000130387210186
      +54.0, -0.00000149844406951
      +64.0, +0.00000000000000000
     +127.0, -0.00000162654178981
     +128.0, +0.00000000000000000

swesterfeld:
Inserting T=float casts makes the function perform better (at least here). It avoids conversions between single precision and double precision values (i.e. cvtsd2ss) which would otherwise be used. So this version
```
static inline float G_GNUC_CONST
fast_log2ff (float value)
{
  union {
    float f;
    int i;
  } float_u;
  float_u.f = value;
  // compute log_2 using float exponent
  const int log_2 = ((float_u.i >> 23) & 255) - 128;
  // replace float exponent
  float_u.i &= ~(255 << 23);
  float_u.i += BSE_FLOAT_BIAS << 23;
  typedef float T;
  T u, x = float_u.f;
  // lolremez --long-double -d 6 -r 1:2 "log(x)/log(2)+1-0.00000184568668708"
  u =         T (-2.5691088815846393966e-2l);
  u = u * x + T (2.7514877034856806734e-1l);
  u = u * x + T (-1.2669182593669424748l);
  u = u * x + T (3.2865287704176774059l);
  u = u * x + T (-5.3419892025067624343l);
  u = u * x + T (6.1129631283200211528l);
  x = u * x + T (-2.040042118396715321l);
  return x + log_2;
}
```
is faster, because all operations are on floats. This costs a bit of precision but the float version (`fast_log2ff`) is faster than using double (`fast_log2fd`) or long double (`fast_log2fl`).
```
$ g++ -std=c++17 -Wall -g -O3 -o l2 l2.cc `pkg-config --cflags --libs spectmorph glib-2.0 bse`
$ l2
log2f: 5.369997 ns/call
fast_log2fl: 7.890487 ns/call
fast_log2fd: 4.662395 ns/call
fast_log2ff: 3.652096 ns/call

prec: fast_log2ff: 4.493532e-06
prec: fast_log2fd: 3.721012e-06
prec: fast_log2fl: 3.691373e-06


$ clang++ -std=c++17 -g -O3 -o l2 l2.cc `pkg-config --cflags --libs spectmorph glib-2.0 bse`
$ l2
log2f: 5.323792 ns/call
fast_log2fl: 7.597113 ns/call
fast_log2fd: 5.201006 ns/call
fast_log2ff: 4.071403 ns/call

prec: fast_log2ff: 4.493532e-06
prec: fast_log2fd: 3.721012e-06
prec: fast_log2fl: 3.691373e-06
```

On the other stuff I mostly agree. If you have use cases in mind (for key tracking or filter frequency modulation it doesn't matter) that need integers k exp2 (k) to be 2^k and you think you want to pay for it with one add-mul, ok. I think relative error is the most important goal here, though. For instance if the key tracking algorithm returns 222 instead of 220, from a muscians point it is as bad as returning 888 instead of 880. Both sound equally wrong, and both have the same relative error (not absolute error).

Applying corrections for fast_log2 (2^k) to yield k  for integer k sounds ok to me. Note that it doesn't fix fast_log2 (7.999999) to be 3, as you patched only the case where the input is equal to or slightly greater than 2^k, not the case where it is slightly smaller.

```
fast_log2fl (7.999999) = 2.999996; log2f (7.999999) = 3.000000
fast_log2fl (8.000000) = 3.000000; log2f (8.000000) = 3.000000
fast_log2fl (8.000001) = 3.000000; log2f (8.000001) = 3.000000
```
This could be fixed by adjusting the linear coeffcient of the remez polynomial, but this would make our worst case error larger, and I think as the result is so close to the perfect value it is probably not worth it. 

As for whether to approximate at all on AMD64: my impression from the benchmarks is that in many cases using one of the approximations would yield sufficient quality faster that exp2f or log2f. On AMD64 especially when using T=float internally.

However, the gain is not dramatic, and maybe we're trying to optimize something with approximations that is not really a performance problem. For instance the LadderFilter (the place where this started) typically only needs one log2 value per note-on. Only portamento would affect this negatively which we do not support at the moment. What I'm trying to say here is: if we use log2f/exp2f and one day we run perf on beast and see than 10% of the CPU usage is spent in exp2f, we could still deal with it at that point in time.

tim-janik:
I just pushed a new version of the 2^x approximation as fast_exp2() to master and removed the old approximations. Commit 7d919e98bb116f98a6b05e56f8a05edb6825692e still has the old functions, and a comparative benchmark.
The approximation is based on the 5th order polynomial with minimal relative error, approximating 2^x within [-0.5...+0.5], generated with [Sollya](http://sollya.gforge.inria.fr/). The result is almost as accurate as the old bse_approx6_exp2(), although fast_exp() uses only float arithmetic. It is also significantly faster than the old functions and the optimized exp2f() variant recently added to glibc and musl ([ARM-software/exp2f](https://github.com/ARM-software/optimized-routines/blob/master/math/exp2f.c)).

~~~~
git checkout 7d919e98bb116f98a6b05e56f8a05edb6825692e && make 
BSE_TEST=slow out/tests/suite1 fast_math_test  fast_math_bench 
  RUN…     fast_math_test
  PASS     fast_math_test
  RUN…     fast_math_bench
  BENCH    fast_exp2          # timing: fastest=0.000392s calls=458605760.0/s diff=0.00000033028512169686 (@0.734155)
  BENCH    arm_exp2f          # timing: fastest=0.000532s calls=338185006.7/s diff=0.00000005979062378536 (@0.958382)
  BENCH    exp2f              # timing: fastest=0.000650s calls=276832337.1/s diff=0.00000005979062378536 (@0.958382)
  BENCH    bse_approx5_exp2   # timing: fastest=0.000768s calls=234464124.9/s diff=0.00000458521965707170 (@0.500000)
  BENCH    bse_approx6_exp2   # timing: fastest=0.000847s calls=212474853.5/s diff=0.00000022838359092781 (@0.500000)
  BENCH    bse_approx7_exp2   # timing: fastest=0.000854s calls=210667974.8/s diff=0.00000000994037496760 (@0.500000)
  PASS     fast_math_bench
~~~~

tim-janik:
I now have a proper fast_log2() approximation. I used Sollya for this, and figured you only get a good approximation for log2(x+1), see here: https://en.wikipedia.org/wiki/Logarithm#Power_series
That allows the constant of the approximation polynomial to be 0.0, which guarantees correct results for 2^int inputs.
Tests and benchmark are now in master (results from a production build):

~~~~
  BENCH    fast_log2          # timing: fastest=0.000185s calls=487621127.8/s diff=0.00000302175265787241 (@0.000015)
  BENCH    log2f              # timing: fastest=0.000342s calls=263339591.6/s diff=0.00000095218742401926 (@0.000010)
~~~~

PR #125, opened 2019-09-11
Resampler cleanups

swesterfeld:
I cleaned up the resampling code:
* the C API was removed
* the implementation no longer relies on blockutils, but does its own SSE/FPU handling
* Resampler2::create() got an additional argument: true means SSE should be used if available; false means the FPU version should always be used
* FPU/SSE resampling is now under test again, I adapted some of the tests

Plugins (LadderFilter) for now can use Resampler2::create (..., false) to force using FPU resampling. As soon as the engine guarantees SSE-alignment of input/output blocks, this can be removed.

The old Makefile.am had more "testresampler" style tests, these could be added.


swesterfeld:
Since this PR wasn't merged yet, I simply force pushed an updated version. Since we have no more C API, I could modernize the C++ API. It should be used like this:
```
  float input[128];
  float output[256];
  Resampler2 resampler (Resampler2::UP, Resampler2::PREC_96DB);
  resampler.process_block (input, 128, output);
```
Note that process_block() still needs a virtual function call (like it used to be), but I did some benchmarking before refactoring the API. The performance difference between virtual function call and inlined resampler code is minimal. Like this, I could move all implementation bits back into `bseresampler.cc`, so `bseresampler.hh` is really clean and doesn't contain any actual resampling code, and `bseresamplerimpl.hh` could be removed.

I also re-added the old testresampler tests from the Makefile.am, these are added using TEST_SLOW, so `out/tests/suite1 --slow` is needed to run the tests. I also re-enabled the standalone testcode, passing args to the standalone main() from suite1 main()
```
$ testresampler perf # OLD
$ out/tests/suite1 --resampler perf # NEW
```
Btw, if you are interested in the impact of the block size on performance, here is upsampling with different block sizes (times in ns/sample) upsampling, PREC48_DB.
```
4 5.1589
8 3.52419
16 3.80419
32 3.079
64 2.48795
128 2.25006
256 2.09908
512 2.019
```

I recommend merging this PR first, then I can update the LadderFilter PR (https://github.com/tim-janik/beast/pull/122) to use the new resampling API, then merge the LadderFilter.

tests/suite1-main.cc
```diff
@@ -188,6 +188,12 @@ main (int argc, char *argv[])
       bench_aida();
       return 0;
     }
+  if (argc >= 2 && argv[1] && std::string ("--resampler") == argv[1])
+    {
+      Bse::init_async (&argc, argv, argv[0], args);
+      extern int test_resampler (int, char **);
```
tim-janik: NEVER inline prototypes, ALWAYS use a header file!
Without any exception.
I've once spent days debugging misbehaviour due to mismatching prototype, never again!

tim-janik:
Your branch is merged, but the --resampler options needs fixing.

tests/suite1-main.cc
swesterfeld: OK, should be fixed once you merge https://github.com/tim-janik/beast/pull/126

PR #128, opened 2019-09-20
Jack driver new api

swesterfeld:
Ok, here is a port of the Jack driver (https://github.com/tim-janik/beast/pull/31) to the new driver API. I tested playback and also stereo-through. This doesn't yet have midi support.

### DIFF
--- a/bse/driver-jack.cc
+++ b/bse/driver-jack.cc
@@ -1,0 +1,850 @@
+// This Source Code Form is licensed MPL-2.0: http://mozilla.org/MPL/2.0
+#include "driver.hh"
+#include "bseengine.hh"
+#include "internal.hh"
+#include "gsldatautils.hh"
+#include "bseblockutils.hh"
+
+#include <unistd.h>
+
+#define JDEBUG(...)     Bse::debug ("jack", __VA_ARGS__)
+
+#if __has_include(<jack/jack.h>)
+#include <jack/jack.h>
+
+using namespace Bse;
+
+namespace { // Anon
+
+/**
+ * This function uses std::copy to copy the n_values of data from ivalues
+ * to ovalues. If a specialized version is available in bseblockutils,
+ * then this - usually faster - version will be used.
+ *
+ * The data in ivalues and ovalues may not overlap.
+ */
+template<class Data> void
+fast_copy (uint        n_values,
+           Data       *ovalues,
+	   const Data *ivalues)
+{
+  copy (ivalues, ivalues + n_values, ovalues);
+}
+
+template<> void
+fast_copy (uint         n_values,
+           float       *ovalues,
+           const float *ivalues)
+{
+  Block::copy (n_values, ovalues, ivalues);
+}
+
+#if 0 // <- avoid unused warning
+template<> void
+fast_copy (uint	         n_values,
+           uint32       *ovalues,
+           const uint32 *ivalues)
+{
+  Block::copy (n_values, ovalues, ivalues);
+}
+#endif
+
+/**
+ * The FrameRingBuffer class implements a ringbuffer for the communication
+ * between two threads. One thread - the producer thread - may only write
+ * data to the ringbuffer. The other thread - the consumer thread - may
+ * only read data from the ringbuffer.
+ *
+ * Given that these two threads only use the appropriate functions, no
+ * other synchronization is required to ensure that the data gets safely
+ * from the producer thread to the consumer thread. However, all operations
+ * that are provided by the ringbuffer are non-blocking, so that you may
+ * need a condition or other synchronization primitive if you want the
+ * producer and/or consumer to block if the ringbuffer is full/empty.
+ *
+ * Implementation: the synchronization between the two threads is only
+ * implemented by two index variables (read_frame_pos and write_frame_pos)
+ * for which atomic integer reads and writes are required. Since the
+ * producer thread only modifies the write_frame_pos and the consumer thread
+ * only modifies the read_frame_pos, no compare-and-swap or similar
+ * operations are needed to avoid concurrent writes.
+ */
+template<class T>
+class FrameRingBuffer {
+  //BIRNET_PRIVATE_COPY (FrameRingBuffer);
+private:
+  vector<vector<T> >  channel_buffer_;
+  std::atomic<int>    atomic_read_frame_pos_;
+  std::atomic<int>    atomic_write_frame_pos_;
+  uint                channel_buffer_size_;       // = n_frames + 1; the extra frame allows us to
+                                                  // see the difference between an empty/full ringbuffer
+  uint                n_channels_;
+public:
+  FrameRingBuffer (uint n_frames = 0,
+		   uint n_channels = 1)
+  {
+    resize (n_frames, n_channels);
+  }
+  /**
+   * Check available read space in the ringbuffer.
+   * This function may only be called from the consumer thread.
+   *
+   * @returns the number of frames that are available for reading
+   */
+  uint
+  get_readable_frames()
+  {
+    int wpos = atomic_write_frame_pos_;
+    int rpos = atomic_read_frame_pos_;
+
+    if (wpos < rpos)		    /* wpos == rpos -> empty ringbuffer */
+      wpos += channel_buffer_size_;
+
+    return wpos - rpos;
+  }
+  /**
+   * Read data from the ringbuffer; if there is not enough data
+   * in the ringbuffer, the function will return the number of frames
+   * that could be read without blocking.
+   *
+   * This function should be called from the consumer thread.
+   *
+   * @returns the number of successfully read frames
+   */
+  uint
+  read (uint    n_frames,
+        T     **frames)
+  {
+    int rpos = atomic_read_frame_pos_;
+    uint can_read = std::min (get_readable_frames(), n_frames);
+
+    uint read1 = std::min (can_read, channel_buffer_size_ - rpos);
+    uint read2 = can_read - read1;
+
+    for (uint ch = 0; ch < n_channels_; ch++)
+      {
+	fast_copy (read1, frames[ch], &channel_buffer_[ch][rpos]);
+	fast_copy (read2, frames[ch] + read1, &channel_buffer_[ch][0]);
+      }
+
+    atomic_read_frame_pos_ = (rpos + can_read) % channel_buffer_size_;
+    return can_read;
+  }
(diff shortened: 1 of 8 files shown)


bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+
+template<> void
+fast_copy (uint         n_values,
+           float       *ovalues,
+           const float *ivalues)
+{
+  Block::copy (n_values, ovalues, ivalues);
+}
+
+#if 0 // <- avoid unused warning
```
tim-janik: Use BSE_USED or BSE_UNUSED to silence the compiler here.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+  jack_client_t *jack_client = connect_jack();
+  if (jack_client)
+    {
+      devices = query_jack_devices (jack_client);
+      disconnect_jack (jack_client);
+    }
+  else
+    {
+      // should we try to generate and show an error message if connecting jack failed?
+    }
```
tim-janik: No, if there's no jack, there are no devices to list. Simple. Stay silent.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+
+      /* the default device is usually the hardware device, so things should work as expected
+       * we could show try to show non-default devices as well, but this could be confusing
+       */
+      if (details.default_device)
+        {
+          Driver::Entry entry;
+          entry.devid = device_name;
+          entry.priority = Driver::JACK;
+          entries.push_back (entry);
```
tim-janik: Two Driver::Entry structures must never have the same priority. That's why we have predefined priority constants for drivers, cards, devices, subdevices. Also, we need the other Entry fields to be filled to allow GUI selection. For now, it doesn't matter too much how to match name/blurb/status fields to extra info, each is just another line in my current UI prototype, but we need *something* to display devices to the user that make them humanly identifyable.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+        }
+    }
+}
+
+/* macro for jack dropout tests - see below */
+#define TEST_DROPOUT() if (unlink ("/tmp/drop") == 0) usleep (1.5 * 1000000. * buffer_frames_ / mix_freq_); /* sleep 1.5 * buffer size */
+
```
tim-janik: I know this is disabled, atm, but please still make this /tmp/bse-dropout (or similar) so we're not stamping over namespaces if this is activated on purpose or by accident.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+ * operations are needed to avoid concurrent writes.
+ */
+template<class T>
+class FrameRingBuffer {
+  //BIRNET_PRIVATE_COPY (FrameRingBuffer);
+private:
+  vector<vector<T> >  channel_buffer_;
+  std::atomic<int>    atomic_read_frame_pos_;
+  std::atomic<int>    atomic_write_frame_pos_;
+  uint                channel_buffer_size_;       // = n_frames + 1; the extra frame allows us to
```
tim-janik: This is C++17, always and unconditionally initialize all struct members that don't have a ctor. If in doubt, just add =0, =false, =EnumType(0).

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+                                                  // see the difference between an empty/full ringbuffer
+  uint                n_channels_;
```
tim-janik: n_channels_ = 0;

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+
+  uint64                        device_read_counter_ = 0;
+  uint64                        device_write_counter_ = 0;
+  int                           device_open_counter_ = 0;
+
+  static int
+  static_process_callback (jack_nframes_t   n_frames,
+                           void            *jack_handle)
+  {
+    return static_cast <JackPcmDriver *> (jack_handle)->process_callback (n_frames);
```
tim-janik: Use a lambda instead of static_*().

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+            range.max = std::max (range.max, port_range.max);
+          }
+      }
+    return range;
+  }
+  static void
+  static_latency_callback (jack_latency_callback_mode_t  mode,
+                           void                         *jack_handle)
+  {
+    static_cast <JackPcmDriver *> (jack_handle)->latency_callback (mode);
```
tim-janik: Use a lambda instead of static_*().

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+        range.max += buffer_frames_;
+
+        for (auto port : input_ports_)
+          jack_port_set_latency_range (port, mode, &range);
+      }
+  }
+  static void
+  static_shutdown_callback (void *jack_handle)
+  {
+    static_cast<JackPcmDriver *> (jack_handle)->shutdown_callback();
```
tim-janik: Use a lambda instead of static_*(), see below.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+          Bse::warning ("JACK driver: output silence init failed: (frames_written != jack->buffer_frames)\n");
+
+      }
+
+    /* activate */
+    if (error == 0)
+      {
+        jack_set_process_callback (jack_client_, static_process_callback, this);
+        jack_set_latency_callback (jack_client_, static_latency_callback, this);
+        jack_on_shutdown (jack_client_, static_shutdown_callback, this);
```
tim-janik: Note that lambdas that don't take context arguments boil down to normal functions that can be used as function pointer. For example, here is how it looks when you replace the static_* wrappers with lambda:

    jack_on_shutdown (jack_client_, [] (void *p) {
      static_cast<JackPcmDriver*> (p)->shutdown_callback();
    }, this); 

This could also be a one liner if you have a wider terminal.

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+  explicit      JackPcmDriver (const String &devid) : PcmDriver (devid) {}
+  static PcmDriverP
+  create (const String &devid)
+  {
+    auto pdriverp = std::make_shared<JackPcmDriver> (devid);
+    return pdriverp;
+  }
+  ~JackPcmDriver()
+  {
+  }
```
tim-janik: The dtor should release all resources, e.g. call close(), either unconditionally, or with `if (jack_client_)`. It may also be executed without open() being called previously, or after a failing open() without close().

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+
+        uint dummy;
+        pcm_latency (&dummy, &dummy);   // debugging only: print latency values
+      }
+    else
+      {
+        disconnect_jack (jack_client_);
+        jack_client_ = nullptr;
+      }
+    JDEBUG ("JACK: opening PCM \"%s\" readupble=%d writable=%d: %s", devid_.c_str(), readable(), writable(), bse_error_blurb (error));
```
tim-janik: Please make this "JACK: %s: ...", devid_, ...
And fix "readable" ;-)

bse/driver-jack.cc
```diff
@@ -0,0 +1,905 @@
...
+      {
+        printed_xruns_ = atomic_xruns_;
+        Bse::printerr ("JACK: %d beast driver xruns\n", printed_xruns_);
+      }
+    /* report jack shutdown */
+    if (is_down_ && !printed_is_down_)
+      {
+        printed_is_down_ = true;
+        Bse::printerr ("JACK: connection to jack server lost\n");
+        Bse::printerr ("JACK:  -> to continue, manually stop playback and restart\n");
```
tim-janik: Please add the devid_ to all printerr calls, I've also fixed this in the ALSA driver.

bse/driver-jack.cc
swesterfeld: There are never two Driver::Entry structures here. There is a loop, so you might expect it, but basically the code picks the only device which it considers to be the "default" device from the devices map. It used generate a list of devices, but I think thats more harmful than it helps, but I left the devices query and generate list code there, in case we want to do it differently.

Jack is not very descriptive. Typically we only get the client name "system" and the port names "playback_1", "playback_2",... and we know that it is a hardware device. So ok, I'm trying to give all that information to the user in my new version, but still, we for instance do *not* get a soundcard name or anything useful. There is a Jack Metadata API which could be used to attach descriptions to clients, but at least on my system, this returns nothing. So really what we have is "system", "is hardware device", "2*in 2*out", nothing really useful for the user.

bse/driver-jack.cc
swesterfeld: Ok, done.

bse/driver-jack.cc
swesterfeld: Of course. Fixed. Seems the jack driver predates C++11.

bse/driver-jack.cc
swesterfeld: Fixed. Interesting, didn't know that this would work.

bse/driver-jack.cc
swesterfeld: Right. Fixed.

bse/driver-jack.cc
swesterfeld: Fixed.

bse/driver-jack.cc
swesterfeld: Fixed.

swesterfeld:
Ok, last commit should have all the fixes you requested.

bse/driver-jack.cc
swesterfeld: Done.

bse/driver-jack.cc
swesterfeld: Ok, removed the else branch.

bse/driver-jack.cc
swesterfeld: Done.

bse/driver-jack.cc
swesterfeld: Done.

bse/driver-jack.cc
swesterfeld: Done.

bse/driver-jack.cc
swesterfeld: Done.

PR #129, opened 2019-09-26
BSE: Jack: add midi driver

swesterfeld:
So here is a midi (input) driver for jack. The audio/midi driver share the same jack client, I documented the assumptions I have to make for the communication of the two in the code. While the audio driver typically only has one option (the default audio device), for midi there can be more than one choice. For example here is the relevant part of `--bse-driver-list` on my system if I configure jackd to do the midi handling - we get one possible device in this case:
```
Available MIDI drivers:
  jack=system:midi_capture_1:    (Input, 01000000)
        JACK Midi "system:midi_capture_1" [Physical: in-hw-2-0-0-UA-25EX-MIDI-1]
        Hardware Midi Input
        Routing via the JACK Audio Connection Kit
        Note: JACK adds latency compared to direct hardware access
```
And the same system using alsa midi bridge setting in cadence (uses a2jmidid):
```
Available MIDI drivers:
  jack=a2j:Midi Through [14] (capture): Midi Through Port-0: (Input, 01000000)
        JACK Midi "a2j:Midi Through [14] (capture): Midi Through Port-0"
        Hardware Midi Input
        Routing via the JACK Audio Connection Kit
        Note: JACK adds latency compared to direct hardware access
  jack=a2j:UA-25EX [24] (capture): UA-25EX MIDI 1: (Input, 01000001)
        JACK Midi "a2j:UA-25EX [24] (capture): UA-25EX MIDI 1"
        Hardware Midi Input
        Routing via the JACK Audio Connection Kit
        Note: JACK adds latency compared to direct hardware access
```
So here we get more than one device entry, and if there were more midi devices we would get even more. In this case we have relatively long device names, but we need both: client name and port name to distinguish between devices.

I also added a comment about auto connect which should really be configurable one day; still even without that the midi driver should be usable.

### DIFF
--- a/bse/driver-jack.cc
+++ b/bse/driver-jack.cc
@@ -697,14 +697,14 @@ public:
         char port_name[port_name_size];
         jack_port_t *port;
 
-        snprintf (port_name, port_name_size, "in_%u", i);
+        snprintf (port_name, port_name_size, "in_%u", i + 1);
         port = jack_port_register (jack_client_, port_name, JACK_DEFAULT_AUDIO_TYPE, JackPortIsInput, 0);
         if (port)
           input_ports_.push_back (port);
         else
           error = Bse::Error::FILE_OPEN_FAILED;
 
-        snprintf (port_name, port_name_size, "out_%u", i);
+        snprintf (port_name, port_name_size, "out_%u", i + 1);
         port = jack_port_register (jack_client_, port_name, JACK_DEFAULT_AUDIO_TYPE, JackPortIsOutput, 0);
         if (port)
           output_ports_.push_back (port);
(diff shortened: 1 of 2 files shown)


swesterfeld:
I believe this is now ready for merging. The midi driver has has the improvements we talked about:
* both audio and midi driver now support open without auto connect (using "no-auto-connect" devid)
* the "no-auto-connect" [audio|midi] device can be selected in the ebeast preferences dialog 
* there is now a 1:N relationship between audio and midi driver, i.e. you can open more than one midi devid at a time

tim-janik:
Thank you for your efforts.
While this had been merged into Beast, Beast development is continued in [Anklang](github.com/tim-janik/anklang/).
Anklang has recently gotten Jack support, based on your PR submission.

swesterfeld:
This PR is about jack midi support which is not in Anklang (you only merged a jack pcm driver).

PR #136, opened 2020-02-05
TESTS: testresampler: provide more information if accuracy test fails

swesterfeld:
Is this an acceptabe solution for making failed resampler tests report details of why the test failed?


tim-janik:
> Is this an acceptabe solution for making failed resampler tests report details of why the test failed?

Not quite, this patch doesn't affect the failing test:
```
    #6 0x5e0d09 in testresampler_check_filter_impl() tests/testresampler.cc:799
```

PR #142, opened 2020-07-04
BlepSynth initial implementation

swesterfeld:
So here is the first version of my "BlepSynth" subtractive synth. Its based on MinBLEP oscillators and a non-linear Moog VCF (LadderVCF). Obviously this is the initial version and there is still plenty of thing to do, I have put some TODO comments in the code. Once the midi API is available, it should be easy to add it to the code.

I am not sure if the cc/hh files live in the right place and are integrated properly. Also missing are the programs I use for testing/development. Two examples:
- designblep.cc is used to design the blep that is stored an data in a cc file
- testblep.cc is used for testing the blep oscillator

So either I should maintain them independently (out of the beast tree), or it should be added to the beast tree. The old bleposc branch had the tools included.

As for stuff to add, besides exporting some more properties, an ADSR for the filter would improve the stuff you can do with it. The next big addition then would be LFOs for the parameters to be able to use the possibilties provided by the code which was written with per-sample modulation.


swesterfeld:
Btw, one issue I ran into while testing was that the project auto deactivation was triggered. So to be able to test, I disabled the function auto_deactivate in bseproject.cc.

tim-janik:
Great!

Looks like a really cool device. I've taken your advice and moved the sources into devices/blepsynth/ and adjusted the commit messages and sources. Under devices/, no dedicated Makefile.mk is needed, the *.cc files are picked up automatically.

I've also cleaned up the includes (e.g. we always use our internal assert_return()), implemented NOTE_ON, NOTE_OFF and ALL_NOTES_OFF.

To address some open issues in the sources:

> TODO: unison_voices property should be an integer property, range 1-16, default 1"
* What is missing here? You can just set min/max and step=1. Later we can add a hint that a number-input might be better than a knob here.

> TODO: cutoff property should have logarithmic scaling
* I need the exact mappings here, e.g. minimum, center, maximum, e.g. 32.7, 523.25, 8372.02?
These are C values taken from: https://www.inspiredacoustics.com/en/MIDI_note_numbers_and_center_frequencies
* Or a function with input range, i.e. the above is roughly: 32.7 * 2**x

> TODO: we need non-linear translation between percent and time/level
* Same here, does the discussion we previously had on panning help here?

> TODO: replace this with true midi input
* Done, can we get rid of the Note toggles now?

> TODO: should be easier to get choice value
* Yes, I agree, we should make a topic on choice -> audio value mapping.

> TODO: under some conditions we could enable SSE in LadderVCF (alignment and block_size) *
* Anything needed for that from my part? We already utility SIMD and auto-vectorization in the production build.

swesterfeld:
Ok, now that this is merged I think we can close the issue, certainly the other issues can be sorted out later, we can't have everything at once.
 
> > TODO: under some conditions we could enable SSE in LadderVCF (alignment and block_size) *
> 
> * Anything needed for that from my part? We already utility SIMD and auto-vectorization in the production build.

No. I'll do that once I have the time.

> > TODO: we need non-linear translation between percent and time/level
> 
> * Same here, does the discussion we previously had on panning help here?

I think we had this working. Note that sustain is still missing, there should be a non-linear mapping for sustain% to level.

PR #143, opened 2020-07-18
Blepsynth improvements

swesterfeld:
Here is a bunch of new stuff for BlepSynth, I won't try to reproduce the commit entries here. Only a few additional observations:

* x^3 non-linear adsr time scaling used to work but now it doesn't probably due to parameter API changes (I didn't try to fix it)
* some parameters (octave, unison voices) need step=1 but I haven't seen how to set it when registering a parameter, so this workarounded a bit by overriding param-to-text conversion and omitting decimals
* new filter envelope is hard-wired to modulate cutoff, whereas it would make sense if it could modulate any parameter, but we don't have the ui for selecting modulation targets yet (nor the infrastructure bits)

### DIFF
--- a/devices/blepsynth/TODO
+++ b/devices/blepsynth/TODO
@@ -1,0 +1,9 @@
+QUALITY:
+- seek: slope approximation could improve initial conditions
+
+PERF:
+- specializations for input channels which are not connected
+- speed up dc computation
+- sync specialization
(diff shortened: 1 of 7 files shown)


tim-janik:
Thanks, its merged now, except for triggering C-G, these still help with testing atm.

PR #144, opened 2020-07-26
Support wave-form plotting for BlepSynth

swesterfeld:
It would be nice if the oscillator wave form (which depends on the oscillator parameters) could be displayed at the UI. I've ported the necessary plotting code from Python (plotblep.py) to C++.

The actual waveform is sampled, so one parameter to tune might be how many samples to use. Or the overlap (copy_samples). Finally we currently don't have the infrastructure to display this, so when enabled, the wave forms of both oscillators are written to "/tmp/plot1" and "/tmp/plot2", just to be able to test it.


swesterfeld:
Just one more comment on this code: I saw that if you use a very small pulse_width and make sync really high then not all pulses will be visible. A higher number of samples would help, but I think we may want to make the minimum pulse width depend on the sync factor (to plot a wider pulse in order to be able to display it properly). But such fine tuning would be easier if the wave form would be displayed.

PR #145, opened 2020-08-16
LiquidSFZ device

swesterfeld:
Here is an implementation of a LiquidSFZ device. The obvious next step would be allowing the user to specify file. Loading might want to use an extra thread, so that it doesn't block anything.  SFZ files can take a long while to load.


swesterfeld:
I've created an Anklang PR here: https://github.com/tim-janik/anklang/pull/7 so I'll close this one.

